PR #299, opened 2014-12-31
Update license and copyright in all files

henriquemoody:
https://gist.github.com/henriquemoody/69d1399ba83d389e0974

The code in the gist above, written in PHP, was used to make most of these changes. It is not 100% reliable, that's why I had to check every file.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 PHP_CodeCoverage
 
-Copyright (c) 2009-2014, Sebastian Bergmann <sebastian@phpunit.de>.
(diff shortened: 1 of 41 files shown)


GrahamCampbell:
Every file has a typo in it now, in every single one of these pull requests. There should not be a `the` before the name of the package, every time @sebastianbergmann.

henriquemoody:
yeah, I started to change those, also I updated the script but @sebastianbergmann was faster than me. But not all pull requests I sent has the `the`.

henriquemoody:
@GrahamCampbell, it was fixed by commit b3c6891

PR #304, opened 2015-01-22
Updated jquery from 1.11.1 to 1.11.2

GrahamCampbell:
Downloaded from http://code.jquery.com/jquery-1.11.2.min.js.


whatthejeff:
Thanks!

PR #308, opened 2015-02-05
Fix operator precedence issue

JeroenDeDauw:


src/CodeCoverage/Report/Node/File.php
```diff
@@ -654,7 +654,7 @@ protected function processFunctions(PHP_Token_Stream $tokens)
     protected function crap($ccn, $coverage)
     {
         if ($coverage == 0) {
-            return (string) pow($ccn, 2) + $ccn;
+            return (string)( pow($ccn, 2) + $ccn );
```
GrahamCampbell: Incorrect spacing. This would be correctly formatted:

```php
return (string) (pow($ccn, 2) + $ccn);
```
```

JeroenDeDauw: Fixed

Issue #302, opened 2015-01-16
Code Coverage over HHVM Driver


slashmili:
I'm trying to get code coverage for php/hacklang code under HHVM but I have some trouble and the executed lines don't tally. 

Apparently in HHVM driver you are using fb_get_code_coverage(Just for the record in hhvm xdebug_get_code_coverage and fb_get_code_coverage are implemented the same way) which the return values are not same as [xdebug_get_code_coverage](http://xdebug.org/docs/code_coverage). [Link to prove that](https://gist.github.com/slashmili/dc4ab93c298193430349)

I did some fix on src/CodeCoverage/Driver/HHVM.php it's not final yet but I did manage to get the proper reports in code coverage(I'll commit it in a bit)

The problem I have now is about code coverage report regarding number of methods. If you look at [this gits](https://gist.github.com/slashmili/8a7c725594e3ec96d834) when I run code coverage on HHVM, only [one method](https://gist.github.com/slashmili/8a7c725594e3ec96d834#file-code-coverage-output-under-hhvm-using-fb_get_code_coverage) is detected but on PHP, code coverage code detects [2 methods](https://gist.github.com/slashmili/8a7c725594e3ec96d834#file-code-coverage-under-xdebug-using-xdebug_get_code_coverage) which is correct. So far I've found the the place that we count number of methods in [PHP_CodeCoverage_Report_Node_File](https://github.com/sebastianbergmann/php-code-coverage/blob/master/src/CodeCoverage/Report/Node/File.php#L266) and I'm looking into that. However I appreciate if you can help me on that method

farzadghanei:
:+1:

sebastianbergmann:
Closing because of  #324 and 672248f49a25fcca46aa6c73612d6731d6b83c4a. Please open a new ticket if the problem persists.

slashmili:
Well the problem still exists because HHVM doesn't support xdebug 100%

```
hhvm --version
HipHop VM 3.7.1 (rel)
Compiler: tags/HHVM-3.7.1-0-g111c76378ac29aa8f25e6bb3c5f15b628e6bbea1
Repo schema: be14a89778f35ca7f7dd4aca7fff0793e8ae3334


hhvm -v Eval.Jit=false bin/phpunit --coverage-html ./coverage -c phpunit.xml src/Tests/ 
PHPUnit 4.6.6 by Sebastian Bergmann and contributors.

[...]

PR #311, opened 2015-02-17
Fix for HHVM code coverage to bring the it close to real #302

slashmili:
fb_get_code_coverage returns the number of time that a line in a file
was executed which is totaly different what standard xdebug
implementation is. This patch bring the number of executed lines close
to what is supposed to be.


sebastianbergmann:
Closing because of  #324 and 672248f49a25fcca46aa6c73612d6731d6b83c4a. Please open a new ticket if the problem persists.

PR #314, opened 2015-03-06
Merging .cov files ignores their filter settings

svycka:
fixes https://github.com/sebastianbergmann/phpcov/issues/18


sebastianbergmann:
This had to be reverted because it caused the whitelist to be ignored for testsuites that use process isolation.

PR #321, opened 2015-04-09
Added no-cache metatags

thedava:
Added no-cache metatags to disable the browser caching of the html coverage report.


GrahamCampbell:
:-1: Surely you should configure your websever to add these headers if you don't want caching. This now limits people to not having caching.

sebastianbergmann:
Thank you, @GrahamCampbell, for pointing this out. I have reverted the patch.

GrahamCampbell:
:beers:

thedava:
I don't see the point in caching the report data. That is something that constantly changes after every new generation.

PR #322, opened 2015-04-10
Adds SebastianBergmann\GlobalState to Filter.php (updates blacklist)

henderjon:
Using these versions:

screen 20shot 20on 202015-04-10 20at 2013 3a15 3a49

`\SebastianBergmann\GlobalState::Blacklist`, `\SebastianBergmann\GlobalState::Restorer`, `\SebastianBergmann\GlobalState::Snapshot` all appear in my coverage report.

screen 20shot 202015-04-10 20at 2013 18 26

I'm not installing it explicitly, so I'm assuming it's a PHPUnit dependency. However, according to the [PHPUnit Manual](https://phpunit.de/manual/current/en/code-coverage-analysis.html) these files should be excluded. 

>>>
By default, a blacklist is used to exclude files from the code coverage report. This blacklist is pre-filled with the source files of PHPUnit and its dependencies. 
>>>

Adding them here removed them from the coverage report.

### DIFF
--- a/src/CodeCoverage/Filter.php
+++ b/src/CodeCoverage/Filter.php
@@ -63,6 +63,9 @@ class PHP_CodeCoverage_Filter
         'SebastianBergmann\Comparator\Comparator' => 1,
         'SebastianBergmann\Exporter\Exporter' => 1,
         'SebastianBergmann\RecursionContext\Context' => 1,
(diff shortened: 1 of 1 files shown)


donatj:
:point_up_2: :+1:

sebastianbergmann:
Closed via d6b5d8bcdf038eafdcace6c47e6138111f16f02f

Issue #251, opened 2014-07-11
Add support for Clover JSON logfile format


sebastianbergmann:
https://confluence.atlassian.com/display/CLOVER/JSON+reference describes the Clover JSON logfile format. If we had that we could get JavaScript-based visualizations such as http://get.carrotsearch.com/foamtree/demo/demos/coverage.html "for free".

Unfortunately, though, FoamTree is not Open Source. But maybe a friendly D3.js expert could help us out?

PR #336, opened 2015-05-10
Add option to emit a very short json summary of the coverage stats.

blockjon:
When I do a code coverage scan across a PHP codebase, I'm really just trying to get some basic information back and I do not prefer to use the current options (PHP serialized, clover etc...) This option allows me to quickly emit a small file with the results of the coverage scan that I'm looking for.

Using this as my PHPUnit log strategy boosts my phpunit runtime performance by a huge margin.

### DIFF
--- a/src/CodeCoverage/Report/JsonOverview.php
+++ b/src/CodeCoverage/Report/JsonOverview.php
@@ -1,0 +1,74 @@
+<?php
+/*
+ * This file is part of the PHP_CodeCoverage package.
+ *
+ * (c) Jonathan Block <block.jon@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Uses json_encode to write a small json overview containing coverage stats.
+ *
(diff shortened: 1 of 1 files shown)


src/CodeCoverage/Report/JsonOverview.php
```diff
@@ -0,0 +1,74 @@
...
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Uses json_encode to write a small json overview containing coverage stats.
+ *
+ * @category   PHP
+ * @package    CodeCoverage
+ * @author     Jonathan Block <block.jon@gmail.com>
+ * @copyright  Jonathan Block <block.jon@gmail.com>
```
blockjon: I just used the comments from other files as a template for this file. I don't need the copyright as such to be part of this file... we can change this if necessary in order to get this to be accepted.

sebastianbergmann:
Thank you for your contribution but I do not think that it would make sense to bundle this. For one thing, adding would at least lead to confusion once #251 is implemented.

PR #337, opened 2015-05-13
Significantly improve merge performance and fix false uncovered lines

javer:
Do not process uncovered files from whitelist while merging several php-coverage dumps and do it only when generating a report.

This also fixes false positive uncovered lines (like single curly brace on a line) for files which are not covered in each php-coverage dump.

Speedup when merging 38 php-coverage dumps is:
php report:  5m15.743s -> 0m1.844s = 171x
html report: 5m57.710s -> 0m23.524s = 15x

### DIFF
--- a/src/CodeCoverage.php
+++ b/src/CodeCoverage.php
@@ -349,7 +349,7 @@ class PHP_CodeCoverage
      */
     public function merge(PHP_CodeCoverage $that)
     {
-        foreach ($that->getData() as $file => $lines) {
+        foreach ($that->getData(true) as $file => $lines) {
             if (!isset($this->data[$file])) {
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
This pull request cannot be automatically merged. Can you please update it? Thanks!

javer:
This PR is obsolete because such fixes have already been applied in https://github.com/sebastianbergmann/php-code-coverage/commit/f15cfc19bfe09e01095afca6793c39881e6622dc. But that commit reintroduced the old bug https://github.com/sebastianbergmann/phpcov/issues/18 which has been fixed in https://github.com/sebastianbergmann/php-code-coverage/pull/343.

PR #338, opened 2015-05-24
disable-code-coverage-ignore

abacaphiliac:
added flag to code coverage class with a tested short circuit statement so that lines to ignore are not calculated regardless of the annotations in the source. by default the ignored lines will be calculated, just as they were before.

@sebastianbergmann i saw your presentations at php[tek] 2015 and thought this was a great idea. what do you think of my implementation? my next step would be to update the phpunit project to utilize this new flag.


sebastianbergmann:
Merged, thanks.

PR #345, opened 2015-06-05
Fix XML coverage report to honor the obviously new data type for $result

theseer:
The current XML version of code coverage report is broken as due to the data type change for $result, all tests are considered skipped ( an (int) cast of an array returns 1 aka SKIPPED as soon as it's not empty).

This PR fixes that.


sebastianbergmann:
Merged manually, thanks.

PR #356, opened 2015-07-08
Removed obsolete isset()

staabm:
`unset` wont do anything in case var/array/offset doesn't exist.


sebastianbergmann:
Cherry-picked into `2.1`, thanks.

PR #359, opened 2015-07-15
drop --dev switch in composer install step

staabm:
`--dev` packages are installed by default. This fixes a composer warning.


sebastianbergmann:
Cherry-picked into `2.1`.

PR #364, opened 2015-07-27
use PHP_CodeCoverage_Driver::* constants instead of magic values.

staabm:
as pointed out in https://github.com/sebastianbergmann/php-code-coverage/commit/48bea1f4aa5fe4eb5547bb578bca39a5b2431915#commitcomment-12368678


sebastianbergmann:
Merged manually, thanks.

PR #368, opened 2015-08-03
Respect BlackListing/WhiteListing of merged .cov files

foozee:
Black Lists and White Lists of merged .cov files should be merged into their counterparts in the current coverage report *before* actually checking if a file from that report exists in a black/white list. in `$this->filter->isFiltered($file)`

### DIFF
--- a/src/CodeCoverage.php
+++ b/src/CodeCoverage.php
@@ -340,7 +340,7 @@ class PHP_CodeCoverage
      */
     public function merge(PHP_CodeCoverage $that)
     {
-       $this->filter->setBlacklistedFiles(
+        $this->filter->setBlacklistedFiles(
             array_merge($this->filter->getBlacklistedFiles(), $that->filter()->getBlacklistedFiles())
         );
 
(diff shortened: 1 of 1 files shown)


src/CodeCoverage.php
```diff
@@ -340,6 +340,14 @@ public function append(array $data, $id = null, $append = true, $linesToBeCovere
      */
     public function merge(PHP_CodeCoverage $that)
     {
+       $this->filter->setBlacklistedFiles(
```
GrahamCampbell: cs

foozee: what does "cs" mean?

GrahamCampbell: you have 7 spaces instead of 8 spaces on this line

foozee: okay, Coding Standards... what could be enhanced here ...

GrahamCampbell: > you have 7 spaces instead of 8 spaces on this line

foozee: okay fixing that, did it quickly on github's editor, seemed a lot simpler than cloning the repo and doing all those things :D 
also sorry, didn't see ur second comment before I put mine

sebastianbergmann:
Merged manually, thanks.

PR #371, opened 2015-08-15
Remove redundant null check

kelunik:
`isset` already checks for `null`.


sebastianbergmann:
Merged manually, thanks.

PR #372, opened 2015-08-15
Fix coverage for nested functions and classes

kelunik:
Some example that fails without this patch:

```php
function foo () {
	function bar () {
		return "foo";
	}

	$b = 14;
}
```

This also affects anonymous classes, they're not recognized and are named `extends` in the reports if there's a `new class extends ...`.

* for functions, `$b = 14` would be marked as executed, but doesn't count towards `foo`'s coverage.
* for anon classes, there's a lot more that breaks, so there's another issue, however, this stacked functionality will be needed, once the other issue is fixed, see #373.


sebastianbergmann:
Merged manually, thanks.

PR #374, opened 2015-08-16
Fix typo...

kelunik:
Seems like I only fixed it in my composer directory and didn't copy it over...


sebastianbergmann:
Merged manually, thanks.

PR #375, opened 2015-08-16
Fix #370: Show 0 / 0 for classes without methods instead of empty cell

### LINKED ISSUE #370: Confusing statistics for source file that declares a class without methods
screenshot from 2015-08-12 09-04-43

screenshot from 2015-08-12 09-05-50

kelunik:


sebastianbergmann:
Merged manually, thanks.

PR #393, opened 2015-10-25
Fix composer install

Maks3w:
Tuning composer.json allow install out of the box

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,6 +42,8 @@
             "src/"
         ]
     },
+    "minimum-stability": "dev",
+    "prefer-stable": true,
     "extra": {
         "branch-alias": {
             "dev-master": "2.2.x-dev"
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ php:
     - 5.6
 
 before_script:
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -42,6 +42,8 @@
             "src/"
         ]
     },
+    "minimum-stability": "dev",
```
GrahamCampbell: why is this needed?

Maks3w: Allow me rewrite your question

Why both settings are needed?

It download the sources when is not possible resolve an stable version (when circular dependencies occur)

GrahamCampbell: no - the question was why isn't stable good enough?

Maks3w: Again. Both settings works in tandem.

https://travis-ci.org/sebastianbergmann/php-code-coverage/jobs/87362951#L102-L149

GrahamCampbell: massive :-1: for this change

GrahamCampbell: your link there answers my question with, no, it's not needed, and shouldn't be added for no reason

sebastianbergmann:
Makes no sense to me. And it works for me without this change.

PR #394, opened 2015-10-27
PHP 7 compatibility

crishoj:
Passing a boolean to `debug_backtrace` causes a  ` TypeError` in PHP 7

### DIFF
--- a/src/CodeCoverage/Exception/InvalidArgumentException.php
+++ b/src/CodeCoverage/Exception/InvalidArgumentException.php
@@ -21,7 +21,7 @@ class PHP_CodeCoverage_InvalidArgumentException extends InvalidArgumentException
      */
     public static function create($argument, $type, $value = null)
     {
-        $stack = debug_backtrace(PHP_VERSION_ID < 50306 ? false : 0);
+        $stack = debug_backtrace(0);
 
         return new self(
             sprintf(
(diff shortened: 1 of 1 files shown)


src/CodeCoverage/Exception/InvalidArgumentException.php
```diff
@@ -21,7 +21,7 @@ class PHP_CodeCoverage_InvalidArgumentException extends InvalidArgumentException
      */
     public static function create($argument, $type, $value = null)
     {
-        $stack = debug_backtrace(false);
+        $stack = debug_backtrace(PHP_VERSION_ID < 50306 ? false : 0);
```
Maks3w: Since master only supports PHP >= 5.6 you can use `0` always

Maks3w: I wonder if you could write `! DEBUG_BACKTRACE_PROVIDE_OBJECT`

crishoj: Thanks for the note on 5.6.

The parameter is a bitmask, so `! DEBUG_BACKTRACE_PROVIDE_OBJECT` wouldn't have the desired effect.

Maks3w:
@sebastianbergmann This can be merged

sebastianbergmann:
Merged manually, thanks.

PR #395, opened 2015-10-27
Initialize after validate

Maks3w:


sebastianbergmann:
What does this solve / achieve?

Maks3w:
Nothing in particular just apply the same CS guideline as 186513dfef8147abbf3cc883cb237e6347660430

sebastianbergmann:
Merged manually, thanks.

PR #396, opened 2015-10-27
Removed obsolete isset()

Maks3w:
Unset wont do anything in case var/array/offset doesn't exist.

Same as 1e4aa0f05ff2d234b306cdc5eb4718b7a18c0fef


GrahamCampbell:
How does calling isset perform vs calling unset when there's nothing to unset?

sebastianbergmann:
Merged manually, thanks.

PR #398, opened 2015-11-03
Tests for Reports

Maks3w:


tests/_files/ignored-lines-crap4j.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<crap_result>
+  <project/>
```
Maks3w: This vary between PHP 5.6.x versions.

sebastianbergmann:
Merged manually, thanks.

PR #400, opened 2015-11-04
Feature/branch coverage

Maks3w:
This PR continue the work @sebastianbergmann made on `feature/path-coverage` branch.

Includes PRs #398 and #399 

Close #380 

TODO:
* Clover XML Report. I won't add support for `cond` line types.

PR #401, opened 2015-11-05
NonStandard pathCovered for clover lines report

Maks3w:
This PR is different from #400 because add a new attribute with the pathCoverage value (covered or not covered) to each line.

This attribute is not part of clover.xsd definition.

I need this attribute per line instead by `cond` statament (as clover.xsd suggest) because I find more useful for print HTML reports line by line.


sebastianbergmann:
Closing #400 and #425 because they have run out-of-sync with `master`. Please coordinate, @Maks3w and @posey2010, and send a new pull request. Thanks!

PR #404, opened 2015-11-06
Fix Xdebug required version

Maks3w:
Composer dev dependency requires Xdebug 2.2.1 https://github.com/sebastianbergmann/php-code-coverage/blob/master/composer.json#L37


sebastianbergmann:
Merged manually, thanks.

PR #405, opened 2015-11-23
Add endLine information of class, trait and function for report output

ccirello:
External tools that have use cases beyound coverage report also need
the endLine of classes, traits and functions to generate whole outputs.

The summing of the startLine and coverageData is not enough to determine
the last line of them, whereas this information has already been
calculated for endLines attribute. Therefore, it comes for free in terms
of computation.


ccirello:
ping @sebastianbergmann

Issue #234, opened 2014-03-30
Optionally raise an exception when a specified unit of code is not executed


sebastianbergmann:
Optionally raise an exception when a unit of code that is specified using either the `@covers` or `@uses` annotation is not executed during a test.

PR #413, opened 2015-12-23
#234: add optional check for unexecuted covered or used code

belanur:
Introduces an optional check that will throw a `PHP_CodeCoverage_CoveredCodeNotExecutedException` if code that is referenced in `@covers` or `@uses` is not executed. 

The check can be activated by calling `PHP_CodeCoverage::setCheckForUnexecutedCoveredCode(true)`.

Please have a look at the unit tests, as I am not exactly sure if they comply with the current test base,
`PHP_CodeCoverageTest::testAppendThrowsExceptionIfCoveredCodeWasNotExecuted()` and `PHP_CodeCoverageTest::testAppendThrowsExceptionIfUsedCodeWasNotExecuted()` in particular.

### DIFF
--- a/src/CodeCoverage/Exception/CoveredCodeNotExecutedException.php
+++ b/src/CodeCoverage/Exception/CoveredCodeNotExecutedException.php
@@ -1,0 +1,18 @@
+<?php
+/*
+ * This file is part of the PHP_CodeCoverage package.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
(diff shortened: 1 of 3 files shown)


belanur:
This fixes #234

src/CodeCoverage/Exception/CoveredCodeNotExecutedException.php
```diff
@@ -0,0 +1,18 @@
...
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Exception that is raised when covered code is not executed.
+ *
+ * @since Class available since Release 2.0.0
```
GrahamCampbell: this line is incorrect

PR #415, opened 2015-12-29
HTML Report: add elapsed time

Slamdunk:
Hi, in our business app the coverage takes more than 2 hours, and needs to be planned precisly.

Adding the total elapsed time in the footer can be a great perk for resource management planning.

I don't know if you accept the new dependency requirement of the [PHP_Timer](https://github.com/sebastianbergmann/php-timer), but it's already installed via [PHPUnit\composer.json#30](https://github.com/sebastianbergmann/phpunit/blob/5.1.3/composer.json#L30)


Slamdunk:
@sebastianbergmann may I ping you for a feedback?

composer.json
```diff
@@ -25,6 +25,7 @@
         "phpunit/php-file-iterator": "~1.3",
         "phpunit/php-token-stream": "~1.3",
         "phpunit/php-text-template": "~1.2",
+        "phpunit/php-timer": ">=1.0.6",
```
GrahamCampbell: Bad version constraint because it has no upper bound.

Slamdunk: I copied it from current [PHPUnit\composer.json](https://github.com/sebastianbergmann/phpunit/blob/5.1/composer.json#L30) from current stable PHPUnit branch, to mantain ubiquity.

What should it be?

sebastianbergmann:
Thank you for your contribution but I do not want this, sorry.

Issue #416, opened 2016-01-08
Xdebug Code Coverage Crashes PHP7


joefallon:
In version 2.1.7 of php-code-coverage the line:
```
xdebug_start_code_coverage(XDEBUG_CC_UNUSED | XDEBUG_CC_DEAD_CODE) 
```
crashes the PHP process with a SIGSEGV in PHP7.

However, if you change the line to read:
```
xdebug_start_code_coverage(XDEBUG_CC_UNUSED)
```
everything is fine and it works normally. 

Also, version 3 removes the blacklist directory feature. Can you bring that back? My code relies on it in a very core way.

GrahamCampbell:
> crashes the PHP process with a SIGSEGV in PHP7.

I've seen this too.

sebastianbergmann:
No, the blacklist will not be brought back. It was intentionally removed. The segfault is an issue that should be reported with Xdebug.

PR #417, opened 2016-01-08
Fixes issue #416 in https://github.com/sebastianbergmann/php-code-cov…

joefallon:
…erage. PHP7 will SIGSEGV if the dead code option is passed in xdebug start code coverage method.


sebastianbergmann:
This needs to be fixed in Xdebug.

PR #419, opened 2016-01-31
Ignore "xdebug://debug-eval(1) : eval()'d code"

Engerim:
The previous implementation does not work in my case because the xdebug file is "xdebug://debug-eval(1) : eval()'d code". With this check now, it is the behaviour from before and allow to exclude my xdebug "file"


Engerim:
ping

sebastianbergmann:
Cherry-picked into `3.1`, thanks.

PR #421, opened 2016-02-11
Fixed bug that causes erroneous uncovered lines under PHPDBG.

ezzatron:
I was attempting to run code coverage under PHP 7 PHPDBG and came across an issue where lines that should appear as covered were appearing as uncovered.

After some investigation, the issue seems to be caused by some strange behaviour in `phpdbg_get_executable()`. I lodged krakjoe/phpdbg#152 to address what I believe to be the cause of this bug, but this PR is a workaround for the current state of affairs.

The issue is that `phpdbg_get_executable()`, when called with the "files" option, can sometimes return information about files other than the ones specified in the "files" option. In addition, when this happens, the "additional" files can sometimes have only a subset of the *actual* executable lines in each file. Because of the way `$fetchedLines` and `$sourceLines` were being merged, executable lines could "disappear" in subsequent calls to `stop()`.

This PR avoids the issue by only adding new data to `$fetchedLines` for the specific files that were requested in the "files" option when calling `phpdbg_get_executable()`.

### DIFF
--- a/src/CodeCoverage/Driver/PHPDBG.php
+++ b/src/CodeCoverage/Driver/PHPDBG.php
@@ -55,6 +55,7 @@ class PHP_CodeCoverage_Driver_PHPDBG implements PHP_CodeCoverage_Driver
 
         if ($fetchedLines == []) {
             $sourceLines = phpdbg_get_executable();
+            $newFiles = array_keys($sourceLines);
         } else {
             $newFiles = array_diff(
                 get_included_files(),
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
I'd rather not have workarounds. @krakjoe, what do you think?

bwoebi:
I fixed the issue in PHP 7.0.4 … so either you wait for the next release in two weeks … or you apply the PR.

sebastianbergmann:
I think I'll wait. Thanks!

sebastianbergmann:
@ezzatron Could you update the pull request to only "repair" the data for PHP < 7.0.4? Thanks!

ezzatron:
No worries, but note that this code will already work for both < 7.0.4 **and** >= 7.0.4.

By "only for < 7.0.4", do you mean putting an explicit `version_compare()` in there? I'm not sure of the motivation for doing so. Performance?

If my understanding of the performance implications of code coverage is correct, this code *should* actually be faster as it has less method calls than the original.

Is there some other reason I'm not seeing?

sebastianbergmann:
I was under the impression that this change is only required for PHP 7.0.1-7.0.3. That is why I would like to have it as a separate execution branch so that it can easily be removed once PHPUnit no longer supports PHP 7.0.

sebastianbergmann:
No feedback, closing.

PR #423, opened 2016-02-18
In order to support PHP 7.0, token stream PHP7 tokens are required

Ocramius:
New tokens were included in `phpunit/php-token-stream:1.4.2` (see https://github.com/sebastianbergmann/php-token-stream/commit/db63be1159c81df649cd0260e30249a586d4129e#diff-07ca42cd43d34e41ef921ccbe0e830deR713)

Without these tokens, this package is not compatible with PHP 7.0, and running coverage will result in a fatal error

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,9 +21,9 @@
         "irc": "irc://irc.freenode.net/phpunit"
     },
     "require": {
-        "php": ">=5.6",
+        "php": "^5.6 || ^7.0",
         "phpunit/php-file-iterator": "~1.3",
-        "phpunit/php-token-stream": "~1.3",
+        "phpunit/php-token-stream": "^1.4.2",
         "phpunit/php-text-template": "~1.2",
         "sebastian/code-unit-reverse-lookup": "~1.0",
         "sebastian/environment": "^1.3.2",


Ocramius:
Note: the problem will likely bubble up to PHPUnit, as phpunit would then (in cascade) need to require the latest version of `sebastianbergmann/php-code-coverage`

sebastianbergmann:
Under which circumstance would installing PHPUnit right now not pull in the latest version of php-code-coverage and php-token-stream?

Ocramius:
@sebastianbergmann this happens for packages testing with `composer update --prefer-lowest --prefer-stable` (mainly eagerly looking for compatibility issues, like indeed this one)

sebastianbergmann:
Thank you, @Ocramius. I have merged this and tagged version 3.2.1 of this package. I would appreciate it if you could investigate whether this indeed "bubbles up" and send me appropriate pull request(s) if it does. I'll try to merge those ASAP. Once again: thanks!

Ocramius:
Indeed, still [bubbles up](https://travis-ci.org/Ocramius/PackageVersions/builds/110060797)

I provided https://github.com/sebastianbergmann/phpunit/pull/2082 as a fix for this. Thanks for the quick tagging!

PR #433, opened 2016-05-03
Speedup

MajorCaiger:
Re-submitted after resolving conflicts as requested.

### DIFF
--- a/src/Driver/Driver.php
+++ b/src/Driver/Driver.php
@@ -39,7 +39,7 @@ interface Driver
     /**
      * Start collection of code coverage information.
      */
-    public function start();
+    public function start($determineUnusedAndDead = true);
 
     /**
      * Stop collection of code coverage information.
--- a/src/Driver/Driver.php
+++ b/src/Driver/Driver.php
@@ -38,6 +38,8 @@ interface Driver
 
     /**
      * Start collection of code coverage information.
+     *
+     * @param bool $determineUnusedAndDead
      */
     public function start($determineUnusedAndDead = true);
 
--- a/src/CodeCoverage.php
+++ b/src/CodeCoverage.php
@@ -114,6 +114,20 @@ class CodeCoverage
     private $unintentionallyCoveredSubclassesWhitelist = [];
 
     /**
+     * Determine if the data has been initialized or not
+     *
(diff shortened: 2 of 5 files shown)


src/Driver/Driver.php
```diff
@@ -39,7 +39,7 @@
     /**
      * Start collection of code coverage information.
      */
-    public function start();
+    public function start($determineUnusedAndDead = true);
```
GrahamCampbell: you forgot to update the phpdoc

GrahamCampbell:
This is a major breaking change due to the modification of an interface. Existing implementations will cause php to fatal error.

GrahamCampbell:
Oh, I see the master is 4.x. That's fine then. ;)

src/Driver/Driver.php
MajorCaiger: Good point, I've added these now

sebastianbergmann:
Thanks, @MajorCaiger. Did you do a benchmark and can share some numbers?

MajorCaiger:
The original benchmarks can be found here https://github.com/sebastianbergmann/php-code-coverage/pull/384 I think the implementation changed slightly but the improvements should be similar.

liorharbater:
Hi @sebastianbergmann 
Using this Speedup enhancement would greatly assist us in reducing our coverage tests cycle time!
Do you plan to release it as part of an official phpunit 4.X release any time soon?
We are using php 5.3 and upgrading to any, more advanced, php version will be too complex at this stage, so we are limited to the 4.X phpunit.

Thanks,
Lior

sebastianbergmann:
There will be no major changes to anything related to PHPUnit 4.8, sorry.

liorharbater:
Unfortunate, but understandable :)
Any plans to release this Speedup enhancement as part of PHPUnit 5.X, any time soon ?

GrahamCampbell:
This is already in PHPUnit 5.x is it not?

willemstuursma:
We've seen a speed increase from 1h48m to 13m. I just want to say I really appreciate this change.

MajorCaiger:
@willemstuursma That's a fantastic improvement!

PR #442, opened 2016-05-31
Introduce start stop

legovaer:
These are changes related to https://github.com/sebastianbergmann/phpcov/pull/51


src/Driver/XdebugSqlite3.php
```diff
@@ -0,0 +1,116 @@
...
+
+namespace SebastianBergmann\CodeCoverage\Driver;
+
+use SebastianBergmann\CodeCoverage\RuntimeException;
+use SebastianBergmann\CodeCoverage\Handler\Sqlite3Data as DataHandler;
+
+/**
+ * Driver for Xdebug's code coverage functionality which saves data to Sqlite3.
+ *
+ * @since Class available since Release 3.0.1
```
GrahamCampbell: might need to be 3.1.0 if we're semver

legovaer: Changed it.

sebastianbergmann:
This is outside the scope of this project, sorry.

PR #464, opened 2016-08-07
Tooltip fuction requires key instead of graph parameter

mithikos:
What I'm using
Generated by php-code-coverage 4.0.1 using PHP 7.0.6 with Xdebug 2.4.1 and PHPUnit 5.5.0 at Sun Aug 7 0:39:01 UTC 2016.

OS: Windows
Browser: Chrome

I get an error  Uncaught TypeError: Cannot read property 'point' of undefined

### DIFF
--- a/src/Report/Html/Renderer/Template/dashboard.html.dist
+++ b/src/Report/Html/Renderer/Template/dashboard.html.dist
@@ -216,7 +216,7 @@ $(document).ready(function() {
       .showLegend(false)
       .forceX([0, 100]);
     chart.tooltipContent(function(key, y, e, graph) {
-      return '<p>' + graph.point.class + '</p>';
+      return '<p>' + key.point.class + '</p>';
     });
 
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
This conflicts with #467.

@mithikos and @garex: Can you have a look at what is currently in `master` and let me know what should be in `master` to make this work? I'm afraid I do not know enough about JavaScript and/or the library in question to fix this. Thanks!

garex:
@sebastianbergmann it's a strange PR. In my previous PR, which is now should be in master, this signature was used:

```
-    chart.tooltipContent(function(key, y, e, graph) {
+    chart.tooltipContent(function(graph) {
```

So how @mithikos created those PR?

Hmm.. This only could be if you not merged my PR?

garex:
As I see in master -- there is already new signature.

May be you did not create a tag and/or did not pushed it to packagist?

Issue #468, opened 2016-08-19
Global functions are not properly counted


Seb35:
I found two issues – I guess bugs – with global functions, mainly in HTML and XML reports (it seems this metric is not reported in other formats – I’m quite new in the world of testing).
- The totals of global functions are always 0/0 even if they individually have the score 1/1, in the file itself and in the parent directory.
- The CRAP is always 0 for global functions; given they are not no-op, the CRAP should be 1 or greater.

I tested with PHPUnit 5.5.2 + PHP_CodeCoverage 4.0.1 with the three drivers: PHP 5.6.24+Xdebug 2.2.5, PHP 7.0.9+Xdebug 2.4.0, PHPDBG 7.0.9. They all report the same figures. Below are the screenshot
![bug_php_codecoverage_directory](https://cloud.githubusercontent.com/assets/772681/17823648/da62cb6e-665d-11e6-93c7-17f98683ac0a.png)
![bug_php_codecoverage_file](https://cloud.githubusercontent.com/assets/772681/17823655/deeda7e4-665d-11e6-9284-5e8a1a5ef9ff.png)

The XML returns, between other things:
`<functions count="1" tested="0" percent="0.00%"/>`
`<function name="a" signature="a($b, $c=false)" start="3" crap="0" executable="3" executed="3" coverage="0"/>`

I searched around existing open and closed issues, I didn’t find such issue. Other have similar effects (total = 0/0) because of no executable lines (#428) or other issues but unrelated to global functions (#422).

Seb35:
I investigated and found the CRAP and coverage (percentage) are not updated in src/Node/File.php, at the contrary of methods. This solved the issue with CRAP and coverage in HTML and XML.

For the totals, only the methods are counted. Similarly to the statistics "Classes and Traits", the statistics "Functions and Methods" should be the addition of the two figures.

I prepared a patch, but I don’t have tests for now. I saw the existing tests only have an empty global function, so the CRAP is really 0 in this case (and its coverage is the issue mentionned elsewhere with 0/0). And the totals are only visible in the XML format – in addition of HTML – right? XML would be more testable than HTML.

Seb35:
[...]

PR #469, opened 2016-08-19
Update the CRAP of global functions and count them in HTML report

Seb35:
This fixes two issues with global functions:
- the CRAP and coverage for global functions was not updated at the end
  of the count; it was only the case for methods, classes, traits.
- in the HTML report, only the methods were counted in the 2nd column
  despite the title "Functions and Methods".

See issue #468


sebastianbergmann:
Merged manually, thanks.

Issue #488, opened 2016-11-29
Executable code not shown as executed/executable


sebastianbergmann:
Moved here from #411 (where more than one issue was discussed).

## Environment
* Linux
* PHP 7.1.0RC6
* Xdebug 2.5.0RC1
* PHPUnit 5.6.7
* php-code-coverage 4.0.3

## Steps to reproduce
```
git clone https://github.com/ramsey/uuid-doctrine.git
cd uuid-doctrine
git checkout abbcbe56b023319dea7a72177333089cf47f5a69
composer install
./vendor/bin/phpunit
```

## Actual result
![e9f95398-d168-11e5-901a-a8b81661e3a2](https://cloud.githubusercontent.com/assets/25218/20706236/9494ee8c-b626-11e6-992d-8001da70cf06.png)

## Expected result
Lines 50-52 of the `uuid-doctrine/src/UuidBinaryType.php` files should not be white.

sebastianbergmann:
I have removed PHPUnit from `require-dev` in the `composer.json` file:
```patch
diff --git a/composer.json b/composer.json
index ee86d64..e8923a0 100644
--- a/composer.json
+++ b/composer.json
@@ -30,12 +30,6 @@
         "ramsey/uuid": "^3.0",
         "doctrine/dbal": "^2.3|^2.5"
     },
-    "require-dev": {
-        "phpunit/phpunit": "^4.7|^5.0",
-        "squizlabs/php_codesniffer": "^2.3",
-        "jakub-onderka/php-parallel-lint": "^0.9.0",
-        "satooshi/php-coveralls": "^0.6.1"
-    },
     "autoload": {
         "psr-4": {"Ramsey\\Uuid\\Doctrine\\": "src/"}
     }
```
This allows me to use a development version of PHPUnit (and its dependencies) which investigating this issue.

sebastianbergmann:
I have added a `print_r()` statement to `src/Driver/Xdebug.php` for printing the raw code coverage data that comes from Xdebug:
```patch
diff --git a/src/Driver/Xdebug.php b/src/Driver/Xdebug.php
index 30099e0..5636297 100644
--- a/src/Driver/Xdebug.php
+++ b/src/Driver/Xdebug.php
@@ -65,6 +65,9 @@ public function start($determineUnusedAndDead = true)
     public function stop()
     {
         $data = xdebug_get_code_coverage();
+
+        print_r($data);
+
         xdebug_stop_code_coverage();
 
[...]

PR #496, opened 2016-12-22
Fix issue #488

JoepRoebroek:
Since I am new to the project and codebase, I am unsure if my fix is correct, but tested it using the method in issue #488 and it works. My concerns are however, as follows: why does this issue occur in PHP7, but not in PHP5?


sebastianbergmann:
There is nothing to fix in #488.

PR #503, opened 2017-01-22
force serialize_precision to ensure test pass with 5.6, 7.0 and 7.1

remicollet:


remicollet:
Bad branch... should be 4.0

PR #504, opened 2017-01-22
force serialize_precision to ensure test pass with 5.6, 7.0 and 7.1

remicollet:
Up to PHP 7.0, "precision" is used (default value is 14)
Since PHP 7.1, "serialize_precision" is used.


remicollet:
Thanks

sebastianbergmann:
Thank you :-)

PR #515, opened 2017-03-31
Fix: wrong use of recursive iterator causing duplicate entries in XML coverage report

theseer:


sebastianbergmann:
Merged manually, thanks!

PR #529, opened 2017-05-18
Suppress warnings when including covered code

### LINKED ISSUE #528: Code coverage is not generated when E_STRICT warnings are produced
It seems that code coverage is not generated if the source code produces `E_STRICT` warnings. I have produced a minimal reproduction case in [joshuaspence/phpunit-test](https://github.com/joshuaspence/phpunit-test).

```
> vendor/bin/phpunit --coverage-xml coverage.xml test/SomeTest.php
PHPUnit 5.7.19 by Sebastian Bergmann and contributors.

Declaration of SomeClass::someMethod($x) should be compatible with SomeOtherClass::someMethod()

> ls
composer.json  composer.lock  phpunit.xml  src  test  vendor
```

joshuaspence:
Fixes #528.

### DIFF
--- a/src/CodeCoverage.php
+++ b/src/CodeCoverage.php
@@ -1080,7 +1080,7 @@ class CodeCoverage
 
             foreach ($this->filter->getWhitelist() as $file) {
                 if ($this->filter->isFile($file)) {
-                    include_once($file);
+                    @include_once($file);
                 }
             }
 


joshuaspence:
I'm happy to write tests for this change, but I'm not quite sure how to go about it.

GrahamCampbell:
👎 If including a file causes a warning, it should be allowed to explode. I don't think this is a bug.

(you can exclude files that you don't want including by coverage)

sebastianbergmann:

PR #539, opened 2017-06-26
Add custom classes for progress bars

### LINKED ISSUE #538: Content Security Issues
| php-code-coverage version | 4.0.8
| PHP version               | 5.6.30
| Xdebug version (if used)  | 2.2.5
| Usage Method              | PHPUnit (via jenkins)
| PHPUnit version (if used) | 5.7.19

Jenkins [applies a content security policy when it serves raw HTML files](https://wiki.jenkins-ci.org/display/JENKINS/Configuring+Content+Security+Policy) - we create a clover HTML file as part of our build and dump the files which are then deployed to the build but when we view the report Firefox and Chrome don't allow the inline styles to work meaning we don't get a particularly helpful dashboard.

[It's a known limitation of Bootstrap's progress bar](http://getbootstrap.com/components/#progress), with the alternatives to do it via JS or use custom classes. I can send the boring PR to add the custom classes and update the template if you think that's the right thing to do?

PatrickRose:
As content security policies are supported across more and more
browsers, the inline styling used for the progress bar widths won't be
processed. Instead of requiring everyone to add a bunch of boring CSP
rules to support this, we add some custom classes for each of the
possible progress bar widths.

We assume that a content security policy allows loading of the
stylesheets, since if it doesn't then there's nothing we can do
anyway.

Fixes #538


sebastianbergmann:
50,005 additions looks extreme to me.

PatrickRose:
It's adding classes for each of the possible percentage points - I can change it to only do classes for the whole numbers?

PatrickRose:
Ah, didn't realise the templates were tested - I'll update those unit tests as well.

sebastianbergmann:
I am wondering whether this change is really worth it. Or rather, whether we are fixing the problem in the right place (or the right problem). The proper solution would be to configure Jenkins with appropriate CSP rules.

PatrickRose:
I'm not sure either tbh. I've never had to interact with CSP before this morning when trying to work out what happened to our clover output!

I think the custom classes approach _is_ the right thing for generated html (though maybe having 100 classes instead of 10000 is better), mainly I think the default CSP setting will always be "Don't allow inline style tags".

stof:
what you could do is to generate the classes only for the values actually used by the report being generated

sebastianbergmann:

Issue #551, opened 2017-09-06
Warning: count(): Parameter must be an array or an object that implements Countable


jtreurniet:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 5.2.2
| PHP version               | 7.2.0RC1
| Driver                    | Xdebug
| Xdebug version (if used)  | 2.6.0
| Installation Method       | PHPUnit PHAR
| Usage Method              | PHPUnit
| PHPUnit version (if used) | 6.3.0

When generating HTML documentation, I'm getting the following warning (the warning is repeated many, many times). The output seems okay, as far as I've been able to tell, however.

```
Warning:  count(): Parameter must be an array or an object that implements Countable in phar:///usr/bin/phpunit/php-code-coverage/Report/Html/Renderer/File.php on line 307
Stack trace:
  1. {main}() /usr/bin/phpunit:0
  2. PHPUnit\TextUI\Command::main() /usr/bin/phpunit:557
  3. PHPUnit\TextUI\Command->run() phar:///usr/bin/phpunit/phpunit/TextUI/Command.php:141
  4. PHPUnit\TextUI\TestRunner->doRun() phar:///usr/bin/phpunit/phpunit/TextUI/Command.php:212
  5. SebastianBergmann\CodeCoverage\Report\Html\Facade->process() phar:///usr/bin/phpunit/phpunit/TextUI/TestRunner.php:598
  6. SebastianBergmann\CodeCoverage\Report\Html\File->render() phar:///usr/bin/phpunit/php-code-coverage/Report/Html/Facade.php:117
  7. SebastianBergmann\CodeCoverage\Report\Html\File->renderSource() phar:///usr/bin/phpunit/php-code-coverage/Report/Html/Renderer/File.php:61
  8. count() phar:///usr/bin/phpunit/php-code-coverage/Report/Html/Renderer/File.php:307
```

If any details are required regarding the environment or configuration I'd be happy to provide those.

remicollet:
Fixed by pr #554, probably worth cherry-pick the fix in 4.0 and 5.0

moon9257:
wordpress 中加载html图片时，报这个错

jdpedrie:
@remicollet any chance of getting this fix cherry-picked over to support PHPUnit 4.8 and 5.7?

sebastianbergmann:
PHPUnit 5.7 is no longer supported.

elyobo:
[...]

PR #553, opened 2017-09-12
Fixes issue #551

pmaselkowski:
The `count` was called on null


src/Report/Html/Renderer/File.php
```diff
@@ -304,7 +304,12 @@ protected function renderSource(FileNode $node)
             $popoverTitle   = '';
 
             if (array_key_exists($i, $coverageData)) {
-                $numTests = count($coverageData[$i]);
+                if(is_array($coverageData[$i]))
```
yergo: these are some serious issues with formatting

pmaselkowski:
@remicollet Thanks for taking care of it :+1:

PR #554, opened 2017-11-03
Fix count(): Parameter must be an array or an object that implements …

remicollet:
…Countable

As `$coverageData[$i]` can be NULL (tested below)

### DIFF
--- a/src/Report/Html/Renderer/File.php
+++ b/src/Report/Html/Renderer/File.php
@@ -304,7 +304,7 @@ class File extends Renderer
             $popoverTitle   = '';
 
             if (array_key_exists($i, $coverageData)) {
-                $numTests = count($coverageData[$i]);
+                $numTests = ($coverageData[$i] ? count($coverageData[$i]) : 0);
 
                 if ($coverageData[$i] === null) {
                     $trClass = ' class="warning"';


sebastianbergmann:
Do you have a reproducing test case that triggers this? I'd rather fix a root cause than a symptom.

remicollet:
From Fedora QA since PHP 7.2.0RC5
https://apps.fedoraproject.org/koschei/package/php-phpunit-php-code-coverage5?collection=f28

```
There was 1 error:
1) SebastianBergmann\CodeCoverage\Report\Html\HTMLTest::testForBankAccountTest
count(): Parameter must be an array or an object that implements Countable
/builddir/build/BUILD/php-code-coverage-8ed1902a57849e117b5651fc1a5c48110946c06b/tests/tests/HTMLTest.php:47
```

Other failure are the same that travis reports.

As you have  `if ($coverageData[$i] === null) { ` I think null is a valid value, but is not coutable in PHP 7.2

remicollet:
> Do you have a reproducing test case that triggers this?

Yes, the test suite included in this repo ;)

sebastianbergmann:
Thanks.

Issue #571, opened 2017-12-12
Feature Request: Ignore @covers option


BackEndTea:
Although the use of  ```@covers``` and ```@coversNothing``` is really useful for generating good coverage reports, there are also times when we want to ignore those settings. For example, i may wish to see exactly what my integration tests are covering.

My proposal would be to add a command line flag to phpunit, for example ```--ignore-covers``` or something like that. Which would cause ```@covers``` and ```@coversNothing``` to be ignored when calculating the code coverage.

Slightly related to: https://github.com/infection/infection/issues/88#issuecomment-351005953

fkupper:
Just came up to the exact same need.

Danon:
I would like to have this option as well, to enable/disable the `@covers` dynamically, depending on needs.

Danon:
You can install tool `rawr/covers-ignore`, which will remove `@covers` and `@coversNothing` annotations from a test suite, which you can then run in `phpunit`.

Here's an example of usage in Github Actions: https://github.com/T-Regx/T-Regx/runs/7260340659?check_suite_focus=true#step:8:1

BrianHenryIE:
I've written a script to match the changed lines in a `git diff` with the covered lines in code coverage output, and run the tests that cover the changed lines.

I use `@covers` everywhere, so unfortunately I'm only getting the tests that are written to test specifically those changed lines. It would be more effective if I could output a second codecoverage report which ignores the `@covers` annotations and gives me the true picture of every test that tested each line.

I'll try #573

sebastianbergmann:
I do not think this would be useful. The purpose of the `#[Covers*]` attributes is to ensure that code coverage is only collected for the code unit(s) that a test is intended to cover. Ignoring these attributes would undermine the very purpose of having them in the first place.

I will not implement this.

danon:
@sebastianbergmann We like using `#[Covers*]` because of the precision it gives us. Obviously using `--ignore-covers` always would defeat the purpuse; but we don't want to disable it **always**, just sometimes, on demand. Sometimes we'd like to run coverage for the sake of normal development, while using `#[Covers*]`. Other times, we'd like to see a "vanilla-coverage" without `#[Covers*]` scoping. Right now there's no way to do it, other than manually remove `#[Covers*]` annotations from codebase.

@sebastianbergmann Do you think you could sleep on that decision some more and take into account this circumstance?

sebastianbergmann:
[...]

PR #573, opened 2018-01-05
Add option to ignore covers annotation

BackEndTea:
This PR:

* Adds the option to ignore coverage annotations, as mentioned in #571 
 
If preferred, i can also make a PR against v5.3 instead

### DIFF
--- a/src/CodeCoverage.php
+++ b/src/CodeCoverage.php
@@ -59,6 +59,11 @@ final class CodeCoverage
     /**
      * @var bool
      */
+    private $ignoreCoversAnnotation = false;
+
+    /**
+     * @var bool
+     */
     private $checkForUnexecutedCoveredCode = false;
 
     /**
@@ -417,6 +422,11 @@ final class CodeCoverage
         $this->forceCoversAnnotation = $flag;
     }
 
+    public function setIgnoreCoversAnnotation(bool $flag): void
+    {
+        $this->ignoreCoversAnnotation = $flag;
+    }
+
     public function setCheckForMissingCoversAnnotation(bool $flag): void
     {
         $this->checkForMissingCoversAnnotation = $flag;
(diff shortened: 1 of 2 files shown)


biganfa:
@BackEndTea why it was closed and not merged? Souds useful. Corresponding stackoverflow question [here](https://stackoverflow.com/questions/48493131/generate-phpunit-code-coverage-while-ignoring-covers/55918286#55918286).

JeroenDeDauw:
Yes, this would be very useful. Right now I can't use the coverage report in several of my projects without either removing the @covers tags or adding a whole lot that I think are better not added.

JeroenDeDauw:
Presumably this was closed because it sat here for one and a half year without any review, even though only a few lines where changed.

kubawerlos:
> Right now I can't use the coverage report

@JeroenDeDauw but you can - the whole point of `@covers*` annotations is to make coverage report more meaningful.

> without either removing the **@Covers** tags or adding a whole lot

I don't understand what's your problem - you can run the PHPUnit with only some test classes having the `@covers*` annotations.

JeroenDeDauw:
@kubawerlos are you saying I can literally add `@covers*` to an integration test to have PHPUnit mark all code it executes as covered?

kubawerlos:
No, I'm saying one way or another - benefit from `@covers*` annotations to know what was intentionally covered or remove them to cover all executed code.

What what you writing I understood you have the annotations, but sometimes you don't want to use them - that's what confuses me - what is the reason of the inconsistency?

JeroenDeDauw:
When you follow something like the clean architecture, you end up with a bunch of use cases that hold application logic and typically each have a request model and a response model. These two model objects are simple DTOs and are not worth unit testing. There are other classes not worth unit testing. The integration tests for the use cases only have one covers tag: for the use case. They do not have a long list of all these other classes, which I think they should not be required to have. So even though my whole application is tested on a sensible level, code coverage shows (incorrectly) low, and is not usable for finding code that is not tested.

Example codebase: https://github.com/wmde/fundraising-donations

fkupper:
Can't you just whitelist/blacklist what you want?

JeroenDeDauw:
That would require many entries and create maintenance overhead. All in all, less preferred than not having the coverage work at all.

PR #588, opened 2018-02-28
Fix creation of the directory for reports

slavcodev:
The attempt to create already existing directory leads to `ErrorException` and not to PHP errors in new PHP, so the suppress errors with `@` is not working.

### DIFF
--- a/src/Report/Clover.php
+++ b/src/Report/Clover.php
@@ -234,7 +234,7 @@ final class Clover
         $buffer = $xmlDocument->saveXML();
 
         if ($target !== null) {
-            if (!@\mkdir(\dirname($target), 0777, true) && !\is_dir(\dirname($target))) {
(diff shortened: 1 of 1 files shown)


sebastianbergmann:

What you claim is not true:
```
php --version
PHP 7.2.2 (cli) (built: Jan 30 2018 10:33:36) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.2.2, Copyright (c) 1999-2018, by Zend Technologies

$ php -a
Interactive shell

php > mkdir('/tmp/foo');
php > mkdir('/tmp/foo');
PHP Warning:  mkdir(): File exists in php shell code on line 1
```
You probably have a custom error handler that converts PHP errors to exceptions.

slavcodev:
Sorry, you're right. Did not notice the framework error handler actions.

PR #602, opened 2018-04-06
Fix: Import exception and add docblock

localheinz:
This PR

* [x] imports an otherwise undefined exception and adds a docblock

Follows https://github.com/sebastianbergmann/php-code-coverage/commit/6d6e40862776020467b22f20f032197c02963097.


localheinz:

PR #615, opened 2018-05-07
Fix PHP Error on mkdir() on existing directory

jasverix:
I cannot do code-coverage because we have a custom error handler that converts PHP Errors into exceptions.

The problem here was that mkdir() was executed before is_dir(). I reversed the direction of these calls (is_dir() before mkdir()) and made them into a function, since the same logic was three times in the class.


sebastianbergmann:

I have just committed e9769cb334f7d0edf5e071ce5183dd8ef0c2464f to address the issue.

PR #619, opened 2018-05-21
Undefined index during code coverage merge

dekker-m:
I'm terrible sorry, but I just noticed our CI server gave an undefined index on line 406 after pull request #614 . I didn't notice it before because it depends on the order in which the files are merged. Yesterday everything seemed fine, but today the test cases spread differently over multiple processes, which resulted in the undefined index errors.

This pull request should fix the undefined index.

PR #621, opened 2018-05-21
Undefined index during merge

dekker-m:
* Fixed undefined index in merge() method (See #619 )
* Removed code from merge() to yield correct merge results
* Added test case for merging coverage files in reverse order (to prevent undefined index again)
* Changed merge test cases to assert against the expected output from a single coverage run, instead of the currently "bugged" result.

### DIFF
--- a/src/CodeCoverage.php
+++ b/src/CodeCoverage.php
@@ -380,30 +380,8 @@ final class CodeCoverage
                 continue;
             }
 
-            if ((\count($lines) > 0) && (\count($this->data[$file]) > 0) && (\count($lines) != \count($this->data[$file]))) {
-                if (\count($lines) > \count($ours = $this->data[$file])) {
-                    // More lines in the one being added in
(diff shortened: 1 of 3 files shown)


dekker-m:
I've just run this against our projects, but the results still don't seem to be correctly. The test cases for php-code-coverage don't seem to cover all edge cases well. I'll check if I can find some of those edge cases within our test data/projects.

dekker-m:
I've been debugging some more tonight. The difference between running test suits in single run or spread over multiple processes are minimal now (< 0.3%). However, the results still don't seem to exactly match. Dead code is merged correctly, multiple test cases covering the same line merge correctly, however there still seems to be an issue with not executed test lines and lines which will never be executed.

I'll let you know once I've figured out some more.

dekker-m:
The latest commit should do the trick, the last few difference were caused by a setting in Codeception

PR #630, opened 2018-07-31
add basic php7.2 compatibility

spipu:
Its allows to add generic Test unit on libraries that are compatible from php5.4 to php7.2


sebastianbergmann:
Version 4 of this component is no longer supported.

PR #641, opened 2018-10-07
Normalize usages of directory separators.

Netmosfera:
Mostly, allowed me to run tests on Win.

### DIFF
--- a/src/Node/AbstractNode.php
+++ b/src/Node/AbstractNode.php
@@ -80,11 +80,7 @@ abstract class AbstractNode implements \Countable
     public function getPath(): string
     {
         if ($this->path === null) {
-            if (
-                $this->parent === null ||
-                $this->parent->getPath() === null ||
-                $this->parent->getPath() === false
-            ) {
+            if ($this->parent === null || $this->parent->getPath() === null || $this->parent->getPath() === false) {
                 $this->path = $this->name;
             } else {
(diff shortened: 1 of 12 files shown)


src/Node/AbstractNode.php
```diff
@@ -80,10 +80,14 @@ public function getId(): string
     public function getPath(): string
     {
         if ($this->path === null) {
-            if ($this->parent === null || $this->parent->getPath() === null || $this->parent->getPath() === false) {
+            if (
+                $this->parent === null ||
```
sebastianbergmann: Please do not mix reformatting like this with what the pull request is actually about.

tests/tests/BuilderTest.php
```diff
@@ -168,45 +181,52 @@ public function testReducePaths($reducedPaths, $commonPath, $paths)
...
-            ],
-            [
+            ];
+
+            yield [
                 [
-                    'Money.php'          => [],
-                    'MoneyBag.php'       => [],
-                    'Cash.phar/Cash.php' => [],
+                    "Money.php"          => [],
```
sebastianbergmann: Please dont replace `'` with `"` here.

Netmosfera: Only that? I thought to maintain the style consistent in that method. The double quotes are not needed in many other places.

sebastianbergmann:
Thanks!

PR #644, opened 2018-10-13
Update bootstrap for HTML coverage to version 4

rpkamp:
Example:

screenshot from 2018-10-13 11-16-05

Previous (bootstrap 3) for this same page for reference:

screenshot from 2018-10-13 12-08-01


src/Report/Html/Facade.php
```diff
@@ -120,35 +120,22 @@ private function copyFiles(string $target): void
...
-                \file_get_contents($this->templatePath . 'css/bootstrap.min.css')
-            )
-
-        );
-
+        \copy($this->templatePath . 'css/bootstrap.min.css', $dir . 'bootstrap.min.css');
         \copy($this->templatePath . 'css/nv.d3.min.css', $dir . 'nv.d3.min.css');
         \copy($this->templatePath . 'css/style.css', $dir . 'style.css');
         \copy($this->templatePath . 'css/custom.css', $dir . 'custom.css');
+        \copy($this->templatePath . 'css/octicons.css', $dir . 'octicons.css');
```
rpkamp: Glyphicons is no longer part of bootstrap, so I opted for [octicons from Github](https://octicons.github.com/) instead because:

1. It's smaller, just some small SVG files instead of large font files
2. The icons are a bit nicer and most programmers should be familiar with them as GitHub uses them as well

src/Report/Html/Renderer/Template/dashboard.html.dist
```diff
@@ -4,23 +4,21 @@
   <meta charset="UTF-8">
   <title>Dashboard for {{full_path}}</title>
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
-  <link href="{{path_to_root}}.css/bootstrap.min.css" rel="stylesheet" type="text/css">
+  <link href="{{path_to_root}}.css/bootstrap.min.css?v=4.1.3" rel="stylesheet" type="text/css">
```
rpkamp: The `?v=` everywhere are used for cache bursting in case people serve this from a web server that's caching (too) aggressively.

rpkamp: What would you like instead? Getting the new HTML with the old CSS / JS is not going to work.

rpkamp: Okay fair enough. I'll remove it.

PR #645, opened 2018-10-19
Fix a crash that can occur when php-token-stream parses invalid files

TysonAndre:
This can happen when an invalid file is in the whitelist or a
whitelisted folder.


sebastianbergmann:
Thanks!

Issue #650, opened 2018-10-29
Wasted screen space in HTML code coverage report


sebastianbergmann:
@rpkamp I suspect this to be a regression from #644. Can you have a look, please?

![screenshot](https://user-images.githubusercontent.com/25218/47644069-7f645480-db6d-11e8-9c04-4118a950dd6f.png)

rpkamp:
No problem, I'll have look somewhere this week :+1:

rpkamp:
I've just checked and I'm not sure about this one. When the code uses the full width but the code itself is small the page looks weird:

![screenshot from 2018-10-31 13-59-05](https://user-images.githubusercontent.com/1059790/47789607-4a930180-dd15-11e8-83fe-14bcc570bc79.png)

So I would suggest one of the following:

1. Make the entire page wider, so both the table at the top as well as the code
1. Add a vertical scrollbar to the code when it's too wide
1. Don't write long lines of code, split them over multiple lines

Also, this is not a regression as the behaviour with bootstrap 3 was the same (see below). But I do agree it could use some love.

![screenshot from 2018-10-31 14-13-39](https://user-images.githubusercontent.com/1059790/47790491-9777d780-dd17-11e8-9d39-df543059da3f.png)

sebastianbergmann:
Thank you for looking into this!

Looks like I was wrong that this is a regression. However, before the Bootstrap update I did not notice this. Maybe it makes it more obvious? I don't know.

Lets agree that this could use some love :-)

Before we try anything else, can you send a pull request that implements your option 1? I think that this might just work.

PR #652, opened 2018-10-31
Use fluid layout for HTML file report

rpkamp:
Possible fix for https://github.com/sebastianbergmann/php-code-coverage/issues/650

Table at the top stays within the viewport, but the code is allowed to flow out of it:
screenshot from 2018-10-31 15-36-37

When the viewport gets smaller the table at the top gets a vertical scrollbar, but the code stays the same:
screenshot from 2018-10-31 15-37-10


sebastianbergmann:
Thanks!

PR #665, opened 2019-02-01
Suggest ext-pcov

fancyweb:


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I feel it is too early to suggest pcov (over Xdebug, for instance).

fancyweb:
No problem and not much time was invested in this change ^^

PR #666, opened 2019-02-01
Also suggests pcov

remicollet:
@krakjoe FYI


sebastianbergmann:
Duplicate of #665.

Please understand that while I merged @krakjoe's pull requests in time for PHPUnit 8, I do think that it is too early to suggest it in `composer.json` or endorse it in any other way than "it is supported in the code".

PR #672, opened 2019-03-06
Do not suggest xdebug extension anymore

andreasschroth:
The suggestion seems biased, e.g. we use pcov now for code coverage which works perfectly fine and is several times faster compared to xdebug.


sebastianbergmann:

PR #677, opened 2019-04-26
Fix: Work around build failures on PHP 7.4

localheinz:
This PR

* [x] works around build failures on PHP 7.4


localheinz:

PR #696, opened 2019-09-02
Trigger build

baspeeters:
Testing to see if builds fail with an arbitrary change, like they do on my Travis account.


baspeeters:
Okay, they don't fail. I'll take it up with TravisCI support.

PR #697, opened 2019-09-02
Throw RuntimeException from CodeCoverage namespace

baspeeters:
Psalm was erroring about exception types in some coverage report formats (in phpunit repo), traced that to here.

It looks like the use of `CodeCoverage\RuntimeException` and `\RuntimeException` is is mixed, making thrown exception types inconsistent.

This PR makes the exceptions consistent. I've touched some places without any coverage, so the changes contain a lot test stuff. Some report facades should now even be 100% covered.

### DIFF
--- a/src/Report/Clover.php
+++ b/src/Report/Clover.php
@@ -9,6 +9,7 @@
  */
 namespace SebastianBergmann\CodeCoverage\Report;
 
+use PHPUnit\Framework\Error\Warning;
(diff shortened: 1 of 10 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it because after a refactoring of how directories are created the changes you proposed were no longer required.

PR #709, opened 2019-11-15
Prioritize pcov above xdebug

Zlender:
Pcov is a much faster code coverage driver and in situtation where both Xdebug and pcov are installed it should be used above xdebug in my opinion.

I've tested with a codebase with some 40k lines of code and 4000 tests.

| Config | Time | Memory |
| ------- | ---- | ---------|
| No coverage| 20s | 280MB|
| Pcov only | 42s | 390MB|
| Pcov and xdebug with patch | 1.44min | 410MB |
| Pcov and xdebug without patch | 8min | 360MB |

It is a change in behavior though as pcov and xdebug might collect different coverage information

### DIFF
--- a/src/CodeCoverage.php
+++ b/src/CodeCoverage.php
@@ -896,12 +896,12 @@ final class CodeCoverage
             return new PHPDBG;
         }
 
-        if ($runtime->hasXdebug()) {
-            return new Xdebug($filter);
-        }
-
         if ($runtime->hasPCOV()) {
             return new PCOV;
+        }
+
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I do not think that there is an issue to solve here: simply do not load Xdebug and PCOV at the same time.

sebastianbergmann:
Changed my mind, thank you for your contribution.

Zlender:
Thank you.

Issue #692, opened 2019-08-15
Code Coverage XML report generate empty files


aviator-ua:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 7.0.7
| PHP version               | 7.3.6
| Driver                    | Xdebug
| Xdebug version (if used)  | 2.7.2
| Installation Method       | Composer
| Usage Method              | PHPUnit
| PHPUnit version (if used) | 8.3.4

Target Class
```php
<?php
declare(strict_types=1);

namespace App\Service;

class SoapClientService
{
    /**
     * @const string
     */
    const URL = 'http://example.com/service.jws?wsdl';

    /**
     * @var \SoapClient
     */
    protected $soapClient = null;

    /**
     * @param string $url
     * @return \SoapClient
     * @throws \InvalidArgumentException
     */
    public function getSoapClient(string $url = null)
    {
        if (!isset($this->soapClient)) {
            try {
                $this->soapClient = new \SoapClient(
                    $url ?? self::URL
                );
            } catch (\Throwable $error) {
                throw new \InvalidArgumentException('Failed to initialize Soap Client');
            }
        }

        return $this->soapClient;
    }
}
```

Target TestCase
```php
<?php
declare(strict_types=1);

namespace App\Tests\Service;

use App\Service\SoapClientService;
use PHPUnit\Framework\TestCase;

/**
 * Unit test for SoapClientService
 */
class SoapClientServiceTest extends TestCase
{

    public function testGetSoapClient(): void
    {
        $client = new SoapClientService();

        $this->expectException(\InvalidArgumentException::class);
        $client->getSoapClient('no-wsdl');
    }
}
```

Command for generate Code Coverage XML Report
```bash
vendor/bin/phpunit -v --coverage-xml build/coverage-xml
```

After executing comand, folder `build/coverage-xml` contains empty files.

Other reports (HTML, Clover, PHP) works as expected.

Also reproduced on the previous version of `sebastianbergmann/php-code-coverage`.

duncan3dc:
Looks like it's just the `SoapClient` construction, here's a smaller reproduction case:
### src/Service.php
```php
namespace Coverage\Example;

class Service
{
    public function __construct()
    {
        new \SoapClient("api/.asmx?wsdl");
    }
}
```

### tests/ServiceTest.php
```php
namespace Coverage\ExampleTests;

class ServiceTest extends \PHPUnit\Framework\TestCase
{
    public function testGet(): void
[...]

PR #721, opened 2020-01-28
Use saveXml() instead of save()

duncan3dc:
As reported in issue #692 there seems to be an issue writing the XML code coverage to disk when using the `\SoapClient` class. This PR gets around that issue by using `saveXml()` to generate the XML and the `file_put_contents()` to write it


sebastianbergmann:
@theseer Do you have an idea what could be going on here? Thanks!

PR #722, opened 2020-02-04
Fix count(): Parameter must be an array or an object that implements Countable

JonasSFreire:
Cherry-pick of #554, in order to add compatibility to php 7.2 for the 4.x release


sebastianbergmann:
There will be no more releases of the 4.x branch.

PR #727, opened 2020-02-12
Fix: Set up PHP with pcov for static code analysis

localheinz:
This PR

* [x] sets up PHP with the `pcov` extension enabled for the static code analysis job

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,14 +28,27 @@ jobs:
       - name: "Checkout"
         uses: "actions/checkout@v2"
 
+      - name: "Install PHP with extensions"
+        uses: "shivammathur/setup-php@v1"
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -24,18 +24,43 @@ jobs:
 
     runs-on: "ubuntu-latest"
 
+    strategy:
+      matrix:
+        php-version:
+          - "7.3"
```
localheinz: Instead of using the matrix here, I could also inline the values.

Would you prefer these values to be inlined?

sebastianbergmann: Static analysis is only done once, with one PHP version. If we do not need a matrix for that: great.

localheinz: 👍

PR #733, opened 2020-02-24
why is $filter not static it is eating RAM

mromot:


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as it makes unrelated changes.

Issue #380, opened 2015-09-01
Support for Path Coverage


mbrodala:
Starting with [version 2.3.0](http://xdebug.org/updates.php#x_2_3_0) Xdebug supports [collecting branch coverage data](http://derickrethans.nl/path-branch-coverage.html).

This package and thus phpunit should use this new feature to gather and output branch coverage in addition to the existing sequence/line coverage.

sebastianbergmann:
This is already being worked on in the https://github.com/sebastianbergmann/php-code-coverage/tree/feature/path-coverage branch. However, I don't think this will be finished anytime soon.

mbrodala:
Good to know that it's already WIP, thanks. :-)

omid1234:
@sebastianbergmann Could we expect any progress here implementing the branch coverage and Dead Code metric in PHPUnit? Please, that is so useful for us and increase the report value of the tool a lot:) Thanks in advance :+1:

sebastianbergmann:
I lack the time to work on this, sorry.

dakorpar:
@sebastianbergmann any status update on this one?

VincentChataignier:
@sebastianbergmann Is this issue still relevant ? Can I submit a PR about this ?
@Maks3w I think you have already worked about branch coverage. Do you have code or PR that I can pick up ?

Thanks :smiley:

sebastianbergmann:
Resource consumption of Xdebug's branch coverage, time and memory usage, is quite high, even for trivial classes and methods. In light of this, I am not sure how useful support for branch coverage in this component would be.

That being said, I would, of course, consider a pull request that implements this.

However, before anyone starts working on an actual implementation: please prototype the visualisation first.

sebastianbergmann:
This visualisation proposed by @derickr is a good start.

![paths-covered-mockup](https://user-images.githubusercontent.com/25218/33260170-5ad42cba-d35f-11e7-8400-dbed33e75902.png)

However, I am not sure how well this will work for real world code with hundreds or thousands (or even more) possible execution paths.

TiMESPLiNTER:
Any updates about path coverage? Either using xdebug or phpdbg?

adam-rocska:
Is this a sick funking joke? Since 2015 there's no progress whatsoever? Even version 9's documentations states "The Opcode Coverage, Branch Coverage, and Path Coverage software metrics are not yet supported by php-code-coverage."... shameful.

[...]

PR #745, opened 2020-05-12
Include driver code in code coverage to enable proper self-test

dvdoug:
Hello @sebastianbergmann

Please see attached a PR to remove `@codeCoverageIgnore` declarations from the driver classes.

Rationale for the change is this:

- In the Xdebug driver, there is code to cleanup bad output from Xdebug (https://github.com/sebastianbergmann/php-code-coverage/blob/master/src/Driver/Xdebug.php#L76)  
- When Xdebug is configured to return branch coverage data (#380), the data that is returned is in a different format from that used by line coverage (see https://xdebug.org/docs/code_coverage)
- This code would therefore need to be adjusted as part of work on #380
- Except that Xdebug returning such bogus data seems like a really weird thing for it to do, and upon investigation I cannot make it do that in my tests
- `git blame` reveals the code has been in the code base since v2 (see https://github.com/sebastianbergmann/php-code-coverage/commit/577fdf3f53ddff207d579f1625fe1398dd2b7613 and https://github.com/sebastianbergmann/php-code-coverage/commit/121d8c7d8c79fd7c77287d2472d1a2dfca1efacf), which combined with the fact I cannot reproduce the behaviour strongly suggests to me that it was to workaround an issue in an ancient version of Xdebug that has long since been fixed
- as a final confirmation to prove the code is now obsolete before submitting a PR to remove it, I wanted to check the code coverage data from php-code-coverage's own test suite - if the code was not executed but the test suite was still passing then that would provide additional reassurance that it could be safely removed
- except that the drivers are excluded from code coverage so it is not possible to use the coverage report to do this

The drivers have been excluded for a very long time, the commit that originally introduced the exclusion for the whole Xdebug driver didn't include any rationale, it simply says "Simplify" (see https://github.com/sebastianbergmann/php-code-coverage/commit/9db2cc428a4cc6152414cfa2de707607d8c190a3). Since then, drivers for phpdbg and pcov have been added but I can't find any reason for the exclusions there other than cargo-culting/copy-pasting.

I can't think of a reason why they should not be included, are you able to remember that far back and why you might have initially excluded them? The phpdbg driver in particular does a lot of post-processing on the data before handing it off and it would be good to have that show up in reports as something that doesn't have tests.

### DIFF
--- a/src/Driver/PCOV.php
+++ b/src/Driver/PCOV.php
@@ -11,8 +11,6 @@ namespace SebastianBergmann\CodeCoverage\Driver;
 
 /**
  * Driver for PCOV code coverage functionality.
- *
- * @codeCoverageIgnore
  */
 final class PCOV implements Driver
 {
--- a/src/Driver/PHPDBG.php
+++ b/src/Driver/PHPDBG.php
@@ -13,8 +13,6 @@ use SebastianBergmann\CodeCoverage\RuntimeException;
 
 /**
  * Driver for PHPDBG's code coverage functionality.
- *
- * @codeCoverageIgnore
  */
 final class PHPDBG implements Driver
 {
--- a/src/Driver/Xdebug.php
+++ b/src/Driver/Xdebug.php
@@ -14,8 +14,6 @@ use SebastianBergmann\CodeCoverage\RuntimeException;
 
 /**
  * Driver for Xdebug's code coverage functionality.
- *
- * @codeCoverageIgnore
  */
 final class Xdebug implements Driver
 {


sebastianbergmann:
@derickr Do you think that https://github.com/sebastianbergmann/php-code-coverage/blob/master/src/Driver/Xdebug.php#L76 can be removed?

sebastianbergmann:
I do not remember the rationale for ignoring the drivers from code coverage.

derickr:
I could never really reproduce Xdebug showing the wrong line numbers, except for a few cases where PHP would add (internally) an extra line at the end of a script. But that's a long time ago, so I think this can be removed.

sebastianbergmann:
Thank you, @derickr.

PR #748, opened 2020-05-13
Encapsulate raw coverage data from drivers

dvdoug:
Hi @sebastianbergmann 

Please find attached a PR for the first piece of the enabling work for #380. The goal of this particular PR is not to add new functionality, merely to lay groundwork.

What does this PR do?
- It replaces the return type of `array` from `Driver->stop()` with a new object `RawCodeCoverageData`, with consequential adjustments to the rest of the code
- It replaces the direct data manipulation that was previously inside `CodeCoverage` with usage of named methods on `RawCodeCoverageData`
- In the Xdebug driver, it adds the ability to accept and propagate driver data in either the traditional line only format, or the extended branch data format

Please note this PR does not adjust the internal `$this->data` array used by the `CodeCoverage` class. I do plan on encapsulating that too, but it will be in a different PR.

### DIFF
--- a/src/RawCodeCoverageData.php
+++ b/src/RawCodeCoverageData.php
@@ -1,0 +1,71 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of phpunit/php-code-coverage.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace SebastianBergmann\CodeCoverage;
+
+/**
+ * Raw context-less code coverage data for SUT.
+ */
+final class RawCodeCoverageData
+{
+    /**
+     * Line coverage data.
+     *
+     * @see https://xdebug.org/docs/code_coverage for format
+     *
+     * @var array
+     */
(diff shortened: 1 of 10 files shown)


sebastianbergmann:
Thanks! I have made some minor adjustments after merging this.

dvdoug:
👍

src/RawCodeCoverageData.php
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * Line coverage data.
+     *
+     * @see https://xdebug.org/docs/code_coverage for format
+     *
+     * @var array
+     */
+    private $lineData = [];
+
+    public function __construct(array $rawCoverage = [])
```
sebastianbergmann, line 26: I do not like the fact that the constructor has to guess the format of the array passed in. Can we please make the constructor private and add separate static methods, named constructors, for each supported format? For instance: `fromXdebugWithoutPathCoverage()`, `fromXdebugWithPathCoverage()`, etc.

dvdoug: Sure thing

sebastianbergmann: Thanks!

PR #749, opened 2020-05-14
Encapsulate processed data

dvdoug:
Hello @sebastianbergmann

This the next piece of work for #380. Like the previous PR, this doesn't add new functionality, it merely reworks some of the existing code to use an object instead of an array.

I've also (as a seperate commit) added in the named constructors you requested in #748


src/RawCodeCoverageData.php
```diff
@@ -21,19 +21,27 @@ final class RawCodeCoverageData
...
-                throw UnknownCoverageDataFormatException::create($file);
-            }
+        foreach ($rawCoverage as $file => $fileCoverageData) {
+            $coverage->lineCoverage[$file] = $fileCoverageData['lines'];
         }
+
+        return $coverage;
+    }
+
+    private function __construct()
```
sebastianbergmann, line 43: Thank you formaking the constructor private and for adopting the concept of named constructors. However, the named constructor must not first create an object and then assign data to it. Instead, the named constructor should pass its parameter, maybe after mapping or processing it, to the constructor.

src/ProcessedCodeCoverageData.php
```diff
@@ -0,0 +1,137 @@
...
+{
+    /**
+     * Line coverage data.
+     * An array of filenames, each having an array of linenumbers, each executable line having an array of testcase ids.
+     *
+     * @var array
+     */
+    private $lineCoverage = [];
+
+    public static function __set_state(array $serialisedData)
```
sebastianbergmann, line 27: Why do you think we need a `__set_state()` method? A value object must not know about persistence or serialization.

dvdoug: It's because of https://github.com/sebastianbergmann/php-code-coverage/blob/master/src/Report/PHP.php#L37

As this is now an object and not an array, the output from `var_export` automatically includes a call to `__set_state` which means it needs to exist as a method for the generated code to continue working.

I could change the PHP export to use `serialize`/`unserialize` instead of `var_export` instead?

sebastianbergmann: Do not worry about the PHP report. We can always bump the major version number and change the output of the PHP report.

src/RawCodeCoverageData.php
dvdoug: Done 🤞

src/ProcessedCodeCoverageData.php
dvdoug: `__set_state` has been removed

Issue #769, opened 2020-06-13
Run tests for each driver


sebastianbergmann:
This library supports

* `phpdbg`
* `php` with `pcov` extension
* `php` with `xdebug` extension version 2
* `php` with `xdebug` extension version 3

In our GitHub Actions CI pipeline, however, we only test with PCOV.

sebastianbergmann:
Maybe @localheinz likes to work on this? :-)

localheinz:
Will take a look this evening!

sebastianbergmann:
@shivammathur Hi! Does `shivammathur/setup-php` have nightly builds of Xdebug 3?

shivammathur:
@sebastianbergmann Yes, for `ubuntu`, Till next release specify `xdebug` in extensions without coverage input and it will enable `Xdebug 3.0.0-dev`. Here is a [test workflow](https://github.com/shivammathur/test-setup-php/actions/runs/137086008/workflow).

sebastianbergmann:
@shivammathur Thanks!

localheinz:
Thank you, @shivammathur!

PR #770, opened 2020-06-13
Enhancement: Run tests with PCOV and Xdebug

localheinz:
This PR

* [x] runs tests with PCOV and Xdebug

Related to #769.

💁‍♂️ Will follow up with more!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,10 @@ jobs:
           - "7.4"
           - "8.0"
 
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Thanks!

sebastianbergmann:
`phpdbg` should be easy and simple as there will be a binary package for it. Xdebug will require us to build the extension ourselves most likely. Not sure if we want to do that.

localheinz:

Issue #774, opened 2020-07-03
Class declaration line is marked as uncovered in PHPUnit 9.3


emodric:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 9.3-dev
| PHP version         | 7.4
| Installation Method | Composer

#### Summary

For some classes, PHPUnit marks the class declaration line (`final class Foo`) as uncovered.

#### Current behavior

![image](https://user-images.githubusercontent.com/362286/86464426-c2f26700-bd2f-11ea-822f-fe58e96dd41b.png)

#### How to reproduce

Not really sure how to reproduce aside from upgrading from PHPUnit 9.2 to 9.3 and migrating coverage config.

#### Expected behavior

To behave as PHPUnit 9.2

![image](https://user-images.githubusercontent.com/362286/86464378-b0782d80-bd2f-11ea-96fa-c4af49a30bc3.png)

sebastianbergmann:
@dvdoug Could this be related to your refactorings / improvements?

dvdoug:
@emodric, which coverage driver are you using? And what value do you have for `processUncoveredFiles`? If the answers are "Xdebug" and "true" then I think this is https://bugs.xdebug.org/view.php?id=1798

emodric:
I use Xdebug 2.9.6 and I do not have `processUncoveredFiles` configured (I don't know the default value).

Note that I only upgraded `phpunit/phpunit` from 9.2 to 9.3 and with it `phpunit/php-code-coverage` from 8.0.2 to `dev-master`. Is this something that manifests only on `dev-master` of `phpunit/php-code-coverage` (due to Xdebug issue you linked)?

dvdoug:
@emodric would you mind deleting `var/cache` and try running your test suite again? I'm curious if that will act as a workaround for you.

The linked Xdebug bug has been there a while I think, but released versions of php-code-coverage seem to accidentally work around it

emodric:
I'm not running this in Symfony context (i.e. no Symfony app, no `var/cache/`, no container). This is a pure unit tests run, no integration testing.

emodric:
To make it clearer, I'm aware that screenshots show that I'm using Symfony controllers to test, that indeed is a run in Symfony context.

BUT, this also happens in a separate run not related to Symfony. I only happened to pick a first class I ran into, which was a Symfony controller :)

emodric:
[...]

PR #775, opened 2020-07-03
Explicitly ignore lines containing Class/Trait/Interface tokens (again)

dvdoug:
For #774 

In currently released versions of php-code-coverage `CodeCoverage->getLinesToBeIgnoredInner()` is called for every file included in the whitelist. This was shared logic between files that were actually processed by a driver and uncovered files that were not. This had the effect that all files passed through our own version of an executable-code detector, which removed lines with class declarations.

In #757, this was refactored so that the inbuilt executable-code detector was only used for uncovered added files. For processed files, the driver's implementation is relied on.

However fixing that issue has revealed https://bugs.xdebug.org/view.php?id=1798 which was previously hidden. Depending on the precise execution path of the test suite, under Xdebug the class declaration in the affected files can show up as (1) non-executable (2) executable but covered by the first test that uses it (3) executable but uncovered. PCOV and PHPDBG are unaffected and always have behaviour (1).

This PR is a small partial revert of #757 to make class declarations always considered non-executable again regardless of what the driver data says.

### DIFF
--- a/src/CodeCoverage.php
+++ b/src/CodeCoverage.php
@@ -563,8 +563,6 @@ final class CodeCoverage
 
                     break;
 
-                 // Intentional fallthrough
-
                 case PHP_Token_INTERFACE::class:
                 case PHP_Token_TRAIT::class:
                 case PHP_Token_CLASS::class:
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
Thank you, @dvdoug.

PR #778, opened 2020-07-13
Additional HTML report views for branch/path data

dvdoug:
Hello @sebastianbergmann 

This is a proposal to add some additional views to the HTML coverage report to better expose branch and path data (when enabled). The changes I'm suggesting are as follows:

1) When branch/path data is available, the directory view is altered so that there are 3 links for each source code file - one for line-based coverage, one for branch coverage and one for path coverage.

image

2) The line-based coverage view is unaltered, except for some minor HTML changes

3) The branch view consists of 2 halves. The first portion is a new version of the familiar coverage view, with each line of source code highlighted in green/red/yellow to indicate coverage with the usual tooltips to show the tests that covered the line. The tooltip will also show the number of branches the line has, and how many have been covered

image

4) Below that, is a second view which shows each individual branch in isolation for ease of identification. I feel this is important as branches and lines can have a many-many relationship, so mentally identifying them from the source view is not always easy.

image

5) The path view works like the branch view does - a version of the traditional view, and then a per-path breakdown.

What do you think?


dvdoug:
🚀

PR #779, opened 2020-07-16
Check for xdebug.mode values from a comma separated string

derickr:
This change is needed so that it is possible to profile PHP running PHPUnit with code coverage


sebastianbergmann:
Merged manually, thanks.

PR #783, opened 2020-07-26
Unconditionally add testCaseId to speed up code coverage massively

derickr:
With this change, both PHP Code Coverage's **and** PHPUnit's test suite still passes.
It makes the PHPUnit test suite run in approximately ⅔rds of the time, with no differences in output:

output-diff

PHPUnit's test suite (`$ php -dxdebug.mode=coverage,profile -dxdebug.start_with_request=yes -dhtml_errors=0 -dmemory_limit=2G ./phpunit --testsuite=unit --coverage-html=/tmp/cov-after.html`):

- Before: `Time: 00:52.085, Memory: 82.00 MB`
- After: `Time: 00:33.547, Memory: 82.00 MB`, a reduction of **36%**.

The valgrind/cachegrind profile shows it is using only **27%** of its previous instructions.

```
/tmp/callgrind.phpunit-coverage.2020-07-25-02-profile-refs.before-pr.out: totals: 178 198 995 103
/tmp/callgrind.phpunit-coverage.2020-07-25-02-profile-refs.out:totals:             48 225 236 059
```

### DIFF
--- a/src/ProcessedCodeCoverageData.php
+++ b/src/ProcessedCodeCoverageData.php
@@ -69,9 +69,7 @@ final class ProcessedCodeCoverageData
         foreach ($executedCode->lineCoverage() as $file => $lines) {
             foreach ($lines as $k => $v) {
                 if ($v === Driver::LINE_EXECUTED) {
-                    if (empty($this->lineCoverage[$file][$k]) || !in_array($testCaseId, $this->lineCoverage[$file][$k], true)) {
-                        $this->lineCoverage[$file][$k][] = $testCaseId;
-                    }
+                    $this->lineCoverage[$file][$k][] = $testCaseId;
                 }
             }
         }


src/ProcessedCodeCoverageData.php
```diff
@@ -69,9 +69,7 @@ public function markCodeAsExecutedByTestCase(string $testCaseId, RawCodeCoverage
         foreach ($executedCode->lineCoverage() as $file => $lines) {
             foreach ($lines as $k => $v) {
                 if ($v === Driver::LINE_EXECUTED) {
-                    if (empty($this->lineCoverage[$file][$k]) || !in_array($testCaseId, $this->lineCoverage[$file][$k], true)) {
-                        $this->lineCoverage[$file][$k][] = $testCaseId;
-                    }
+                    $this->lineCoverage[$file][$k][] = $testCaseId;
```
dvdoug, line 72: If this is safe, then the checks on 83/91 (pre-patch) 81/89 (post-patch) I think should be safe to remove as well, the principle behind adding those was cargo-culting over from this one

exussum12: Not sure what happens in this case but it's possible for the same test to be added to this array twice now.

If the array was keyed by the test id, you could set the key to be the id and the value to be true ? Then use array_keys before processing. 

Either way that's a huge improvement

PR #791, opened 2020-08-10
CacheWarmer: call all available methods for caching

Slamdunk:


src/StaticAnalysis/CacheWarmer.php
```diff
@@ -29,8 +29,16 @@ public function warmCache(string $cacheDirectory, bool $useAnnotationsForIgnorin
         );
 
         foreach ($filter->files() as $file) {
+            /* @noinspection UnusedFunctionResultInspection */
+            $coveredFileAnalyser->classesIn($file);
```
Slamdunk, line 33: These are needed too because cache are per-method:

https://github.com/sebastianbergmann/php-code-coverage/blob/9871cf421031c8a103e246b0bcf460a317d95174/src/StaticAnalysis/Cache.php#L85-L88

sebastianbergmann:
Thank you, Filippo!

PR #797, opened 2020-08-19
fixed namespace handling from removing more than the tail.

MeRoBo:
namespace handling is stripping more than the tail, causing 

unserialize(): Error at offset 209854 of 284023 bytes

[ 
"name" => "Daemon", 
"namespacedName" => "Foo\Infrastructure\Daemon\Watchdog\Daemon", 
"namespace" => "Foo\Infrastructure\\Watchdog";
...]

### DIFF
--- a/tests/_files/ClassThatUsesAnonymousClass.php
+++ b/tests/_files/ClassThatUsesAnonymousClass.php
@@ -1,5 +1,5 @@
 <?php declare(strict_types=1);
-namespace SebastianBergmann\CodeCoverage\TestFixture;
+namespace SebastianBergmann\CodeCoverage\ClassThatUsesAnonymousClass\TestFixture;
 
 final class ClassThatUsesAnonymousClass
 {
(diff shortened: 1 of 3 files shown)


tests/_files/ClassThatUsesAnonymousClass.php
```diff
@@ -1,5 +1,5 @@
 <?php declare(strict_types=1);
-namespace SebastianBergmann\CodeCoverage\TestFixture;
+namespace SebastianBergmann\CodeCoverage\ClassThatUsesAnonymousClass\TestFixture;
```
sebastianbergmann, line 2: Please avoid changing existing test fixture.

sebastianbergmann:
> namespace handling is stripping more than the tail

To make this explicit: when the name of the class is part of the namespace name then it is wrongly removed from the namespace name. Good catch! This is a bug that should be fixed.

> unserialize(): Error at offset 209854 of 284023 bytes

However, I do not understand how a wrong namespace name results in an `unserialize()` error.

MeRoBo:
Hi Sebastianbergmann,

The unserialize doesn’t like the double \ in the namespace value, that is
what brought it to our attention. Unfortunately git hub removed the double
\ in the request.

Thanks for the merge.

PR #805, opened 2020-08-28
Force empty lines non executable

### LINKED ISSUE #799: Uncovered new line at end of file
| php-code-coverage version | 9.1.4
| PHP version               | 7.4
| Xdebug version | 2.9.6
| PHPUnit version | 9.3.7

Since upgrading from PHPUnit `9.2`, new lines at the end of files are being reported as uncovered in the XML report, e.g.
```
<line num="32" type="stmt" count="0"/>
```

It seems to be caused by the removal of this workaround (as reinstating it fixes the clover report): https://github.com/sebastianbergmann/php-code-coverage/commit/d9c9b7b62bbfdd2c503f96fc3af831b8a8cba191

The report is being consumed by SonarCloud - should the last line of files be ignored regardless of the uploaded XML report? If that's the case, I can raise an issue with SonarCloud.

dvdoug:
To hopefully fix #799

This isn't quite the same concept as an "ignored line" so I've deliberately kept this implementation distinct from the static analysis code. It's done as a single pass at the end of execution for speed, although it also simplifies the code.

### DIFF
--- a/src/ProcessedCodeCoverageData.php
+++ b/src/ProcessedCodeCoverageData.php
@@ -188,6 +188,34 @@ final class ProcessedCodeCoverageData
         }
     }
 
+    public function finalize(): void
+    {
+        $this->ensureEmptyLinesAreMarkedNonExecutable();
+    }
+
+    /**
+     * At the end of a file, the PHP interpreter always sees an implicit return. Where this occurs in a file that has
+     * e.g. a class definition, that line cannot be invoked from a test and results in confusing coverage. This engine
+     * implementation detail therefore needs to be masked which is done here by simply ensuring that all empty lines
+     * are considered non-executable for coverage purposes.
+     *
+     * @see https://github.com/sebastianbergmann/php-code-coverage/issues/799
+     */
+    private function ensureEmptyLinesAreMarkedNonExecutable(): void
+    {
+        foreach ($this->lineCoverage as $filename => $coverage) {
+            if (is_file($filename)) {
+                $sourceLines = explode("\n", file_get_contents($filename));
+
+                foreach ($coverage as $line => $lineCoverage) {
+                    if (is_array($lineCoverage) && trim($sourceLines[$line - 1]) === '') {
+                        $this->lineCoverage[$filename][$line] = null;
+                    }
+                }
+            }
+        }
+    }
+
     /**
      * Determine the priority for a line.
      *
--- a/src/CodeCoverage.php
+++ b/src/CodeCoverage.php
@@ -162,6 +162,8 @@ final class CodeCoverage
      */
     public function getData(bool $raw = false): ProcessedCodeCoverageData
     {
+        $this->data->finalize();
+
         if (!$raw) {
             if ($this->processUncoveredFiles) {
                 $this->processUncoveredFilesFromFilter();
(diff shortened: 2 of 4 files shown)


src/ProcessedCodeCoverageData.php
```diff
@@ -188,6 +188,34 @@ public function merge(self $newData): void
...
+     * @see https://github.com/sebastianbergmann/php-code-coverage/issues/799
+     */
+    private function ensureEmptyLinesAreMarkedNonExecutable(): void
+    {
+        foreach ($this->lineCoverage as $filename => $coverage) {
+            if (is_file($filename)) {
+                $sourceLines = explode("\n", file_get_contents($filename));
+
+                foreach ($coverage as $line => $lineCoverage) {
+                    if (is_array($lineCoverage) && trim($sourceLines[$line - 1]) === '') {
```
dvdoug, line 211: The `is_array` isn't technically necessary, without it empty lines that are already non-executable are simply marked non-executable again. However I found it helpful for debugging to ensure that this only triggered on the cases I was expecting it too

sebastianbergmann:
> I've deliberately kept this implementation distinct from the static analysis code. It's done as a single pass at the end of execution for speed, although it also simplifies the code.

The downside of keeping this separate is that the result of this analysis does not get cached.

lcobucci:
@dvdoug @sebastianbergmann it looks like this is didn't solve the issue (or perhaps I'm doing something wrong?): https://github.com/lcobucci/jwt/pull/380

sebastianbergmann:
I am also seeing similar failures:

* https://github.com/sebastianbergmann/code-unit/runs/1061309910
* https://github.com/sebastianbergmann/comparator/runs/1061311241
* https://github.com/sebastianbergmann/global-state/runs/1061313373
* https://github.com/sebastianbergmann/object-enumerator/runs/1061314622
* https://github.com/sebastianbergmann/object-reflector/runs/1061315150
* https://github.com/sebastianbergmann/php-invoker/runs/1061316251
* https://github.com/sebastianbergmann/diff/runs/1061311774
* https://github.com/sebastianbergmann/php-code-coverage/pull/805
* https://github.com/sebastianbergmann/cli-parser/runs/1061320703
* https://github.com/sebastianbergmann/lines-of-code/runs/1061319776
* https://github.com/sebastianbergmann/php-text-template/runs/1061318598
* https://github.com/sebastianbergmann/recursion-context/runs/1061317478
* https://github.com/sebastianbergmann/type/runs/1061317915
* https://github.com/sebastianbergmann/php-timer/runs/1061317036

lcobucci:
@sebastianbergmann in v9.3.4 this component was relying on the driver to detect "dead code" (https://github.com/sebastianbergmann/php-code-coverage/compare/9.1.4...master#diff-b2fbc27b0e544bbffcdfba89b3fb1dd4L640), can the removal of that bit be cause of the problem?

dvdoug:
The issue of the executable last line was originally raised in the context of reporting being incorrect, hence the implemention decision that:

>It's done as a single pass at the end of execution for speed, although it also simplifies the code.

Those CI runs would appear to suggest that strict covers annotations meanss that it's not OK to do the analysis and adjustment as a one-off after the suite completes but requires it to be performed after each individual test invocation instead? That's annoying and will definitely want some caching 😢 

I am happy to look into this more, but please note I am not likely to have time before next week

sebastianbergmann:
> I am happy to look into this more, but please note I am not likely to have time before next week

No worries, take your time. I will look into reverting the recent changes and roll a new release later today.

PR #807, opened 2020-09-05
Empty lines not executable v2

dvdoug:
#799, version 2

This version moves the skip logic from the very end of the process to the very beginning 🤞 that should cover all potential scenarios. This has been tested on #799 (no line 14) and sebastian/code-unit (no risky tests).

This is using a static array cache rather than the static analysis file cache because the file cache is predicated around filenames and methods, but the problem in #799 is lines that are completely outside of a method so those cache keys don't fit the usecase. The analysis itself is also fairly cheap so there's less benefit from persisting that across runs than the more advanced PHP-Parser based analyses so I think that should be OK.

### DIFF
--- a/src/CodeCoverage.php
+++ b/src/CodeCoverage.php
@@ -109,13 +109,6 @@ final class CodeCoverage
     private $parentClassesExcludedFromUnintentionallyCoveredCodeCheck = [];
 
     /**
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
Thanks!

PR #815, opened 2020-09-30
Line coverage of methods should reflect actual coverage

bastien-phi:
Corrects line coverage of methods in Cobertura report.

### DIFF
--- a/src/Report/Cobertura.php
+++ b/src/Report/Cobertura.php
@@ -12,7 +12,6 @@ namespace SebastianBergmann\CodeCoverage\Report;
 use function count;
 use function dirname;
 use function file_put_contents;
-use function max;
 use function range;
 use function time;
 use DOMImplementation;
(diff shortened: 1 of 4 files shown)


bastien-phi:
It looks like decreasing number of lines in `Cobertura.php` also decreases global coverage. Do I need to do something else ?

sebastianbergmann:
CC @smmccabe

smmccabe:
This PR looks good to me, when i tested with gitlab it only reads the stand-alone line reports, not the ones within methods so it wasn't relevant to me. I doubled checked some sample examples and @bastien-phi fix is correct. I based my initial work off of clover which doesn't duplicate the lines, which was incorrect.

tl;dr  - looks good to merge in, thnx @bastien-phi

sebastianbergmann:
Thank you for the review, @smmccabe.

PR #816, opened 2020-09-30
Add endLine in function coverage data

bastien-phi:
Coverage data for function does not have `'endLine'` but data is available. Adding it will be convenient to implement cevrage for non-class based line in Cobertura format (see https://github.com/sebastianbergmann/php-code-coverage/pull/812#issue-490624752)

### DIFF
--- a/src/Node/File.php
+++ b/src/Node/File.php
@@ -532,6 +532,7 @@ final class File extends AbstractNode
                 'namespace'          => $function['namespace'],
                 'signature'          => $function['signature'],
                 'startLine'          => $function['startLine'],
+                'endLine'            => $function['endLine'],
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as by itself this change does not make any sense to me. You can make this change as part of a larger pull request that implements support for the Cobertura XML report for code in the global scope.

bastien-phi:
No problem. 

I thought it made sense to split the two features across different pull requests.

sebastianbergmann:
I agree, although in this case it somehow feels wrong. Sorry.

PR #821, opened 2020-10-08
count on null php7 fix for #ef7b2f5 4.0.8

MockingMagician:
Hello there,

If we refer to the infos the version of phpunit 5.* must be compatible with php 7.* Outside phpunit 5.* uses the version php-code-coverage 4.* which has a typo at line 307 of the file src/Report/Html/Renderer/File.php.

Used, in my case with php 7.4, this error occurs :

> Generating code coverage report in HTML format ...PHP Warning: count(): Parameter must be an array or an object that implements Countable in /mnt/data/files/dev/wam/casting-kingdom/vendor/phpunit/php-code-coverage/src/Report/Html/Renderer/File.php online 307

Because $coverageData[$i] contains null is not countable .

Here is a fix, functional and tested to release a nice version 4.0.9 compatible ...

Thanks in advance :)

### DIFF
--- a/src/Report/Html/Renderer/File.php
+++ b/src/Report/Html/Renderer/File.php
@@ -304,7 +304,7 @@ class File extends Renderer
             $popoverTitle   = '';
 
             if (array_key_exists($i, $coverageData)) {
-                $numTests = count($coverageData[$i]);
+                $numTests = is_null($coverageData[$i]) ? 0 : count($coverageData[$i]);
 
                 if ($coverageData[$i] === null) {
                     $trClass = ' class="warning"';


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it because PHPUnit 5 is no longer supported and neither is the version of php-code-coverage used by PHPUnit 5.

MockingMagician:
Hello @sebastianbergmann 

I understand that PHPUnit 5 is no longer supported, however this is a significant issue in terms of compatibility.

Only one of my current projects is using PHPUnit 5, and this project should be compatible with PHP 5.6 and 7.* but I don't think I'm the only one with this kind of config for a project.

So I find it sad not to fix this, it's really a minor modification (5 min to merge and tag) and ensures compatibility across versions as defined in the package config (composer.json).

Thanks for your feedback.

dvdoug:
@MockingMagician Most projects I've seen will do something like `"phpunit/phpunit": "^5.7||^6.0"` in composer.json in a scenario like yours where they need a newer PHPUnit for newer PHP. Does that work for you?

MockingMagician:
@dvdoug Yes, that's certainly what I'm going to fall back on. I always prefer when the API remains the same whatever the scope of the project, it avoids unpleasant surprises.

It forces me to ask myself other questions, can the xml config be the same, so do I have to maintain two different versions? Can the test API behave differently? And so on and so forth...

I'm just thinking that for a line of code to change and that would allow to respect the very definition that has been given to the package namely : 

- PHPUnit 5 compatible with php 7
- PHPUnit 5 uses phpunit-code-coverage 4 which by extension must be compatible with php 7.

I just think it's a shame to be missing out...

Anyway, thank you both for your feedback. :+1:  @sebastianbergmann @dvdoug

Issue #831, opened 2020-11-23
Files that do not contain a newline are not handled correctly


thirsch:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 9.2.3
| PHP version               | 7.4.12
| Driver                    | Xdebug
| Xdebug version (if used)  | 2.9.8
| Installation Method       | Composer
| Usage Method              | PHPUnit
| PHPUnit version (if used) | 9.4.3

If there is an ~completely empty~ file with only one line without a trailing linebreak in your coverage folder, the coverage can't be created, because the value for $linesOfCode after counting `\n` is 0, whereas the calculated amount of comments is 1 and results in an exception of type SebastianBergmann\LinesOfCode\NegativeValueException.

```
Fatal error: Uncaught SebastianBergmann\LinesOfCode\NegativeValueException: $nonCommentLinesOfCode must not be negative in /.../vendor/phpunit/phpunit/src/TextUI/Command.php on line 103

PHPUnit\TextUI\Exception: $nonCommentLinesOfCode must not be negative in /.../vendor/phpunit/phpunit/src/TextUI/Command.php on line 103

Call Stack:
    0.0006     411104   1. {main}() /.../vendor/phpunit/phpunit/phpunit:0
    0.0162    1110472   2. PHPUnit\TextUI\Command::main() /var/ftproot/www/symfony-tests/vendor/phpunit/phpunit/phpunit:61
```

We've had no issues with PHPUnit 8.* and below with empty files.

PR #832, opened 2020-11-23
Added a test file and testcase for issue #831.

thirsch:
This test shows the issue and also the expected result for #831.


thirsch:
Well, an easy fix would be to simply add one to the line-count here https://github.com/sebastianbergmann/php-code-coverage/blob/master/src/StaticAnalysis/ParsingCoveredFileAnalyser.php#L121

What do you think?

PR #833, opened 2020-11-25
Add CodeCoverage interface.

drupol:
The PHPSpec extension [friends-of-phpspec/phpspec-code-coverage](https://packagist.org/packages/friends-of-phpspec/phpspec-code-coverage) is currently busy [implementing support for PHP8/PHPUnit 9](https://github.com/friends-of-phpspec/phpspec-code-coverage/pull/39).

The [CodeCoverage](https://github.com/sebastianbergmann/php-code-coverage/blob/master/src/CodeCoverage.php) class is now set as final and we are unable to mock it in the tests.

Adding an interface to that class would help to mock it, as suggested by many people [here](https://github.com/phpspec/prophecy/issues/197).

This PR:

* [x] Provides a simple interface for mocking CodeCoverage based on its public methods.

Open questions:

* Should we get rid of the PHPDoc from the CodeCoverage class and move it to the interface only ?

### DIFF
--- a/src/CodeCoverage.php
+++ b/src/CodeCoverage.php
@@ -40,7 +40,7 @@ use SebastianBergmann\CodeUnitReverseLookup\Wizard;
 /**
  * Provides collection functionality for PHP code coverage information.
  */
-final class CodeCoverage
+final class CodeCoverage implements CodeCoverageInterface
 {
     private const UNCOVERED_FILES = 'UNCOVERED_FILES';
 
--- a/src/CodeCoverageInterface.php
+++ b/src/CodeCoverageInterface.php
@@ -15,7 +15,7 @@ use SebastianBergmann\CodeCoverage\Node\Directory;
 
 interface CodeCoverageInterface
 {
-   /**
+    /**
      * Returns the code coverage information as a graph of node objects.
      */
     public function getReport(): Directory;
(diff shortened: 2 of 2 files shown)


sebastianbergmann:
Please do not start discussions such as this one by creating a pull request.

I do not have time at the moment to look into this more deeply, but here are my initial thoughts: I do not like the name `CodeCoverageInterface`, but more importantly I doubt that introducing such an interface right now is the right solution in the long run. While it may solve the problem you want to solve, I do not think that it improves the overall design of this library. Recently @dvdoug invested a lot of time and effort to clean up quite a mess under the hood / in the internals of this library. On the surface, though, this library is still a mess. The public API certainly improved a bit (but not much) since this library was extracted from PHPUnit's code over a decade ago.

That being said, the domain of this library is as technical as it gets. I am not convinced that even if an `CodeCoverageInterface` interface that you can stub exists that configuring these stubs will be convenient or even the right approach to test what you want to test.

FWIW, PHPUnit's test suite does not have tests for PHPUnit's use of this library. PHPUnit starts code coverage data collection before a test, stops code coverage data collection after a test, and retrieves code coverage data at the end and generates reports. Maybe integration tests for this interplay of PHPUnit and php-code-coverage would be nice, but quite frankly I cannot remember a situation in which I missed having such tests.

My suggestion to you would be to try whether stubbing `SebastianBergmann\CodeCoverage\Driver\Driver` works for you. This is what we do in the test suite of this library for testing `SebastianBergmann\CodeCoverage\CodeCoverage`, for instance.

drupol:
Hello,

Thanks for the fast and complete reply.

I though about stubbing the Driver class as well, I will investigate.

Closing this in the meantime.

PR #837, opened 2020-12-02
Allow version 4 of php-token-stream

leofeyer:
Since PHPUnit 8.5 is supposed to work on both PHP 7 and PHP 8, I think we need to allow version 4 of phpunit/php-token-stream to prevent the following error:

```
Error: Class "PHP_Token_NAME_QUALIFIED" not found
```

Support for `NAME_QUALIFIED` has only been added in [version 4.0.4](https://github.com/sebastianbergmann/php-token-stream/blob/master/ChangeLog.md#added), however, this version is not allowed here yet.

Another possible solution would be to allow php-code-coverage version 8 in PHPUnit 8.5, but I figured that this change might have less side-effects. @sebastianbergmann If you prefer the other solution, I'll happily create a new PR. ;)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "ext-dom": "*",
         "ext-xmlwriter": "*",
         "phpunit/php-file-iterator": "^2.0.2",
-        "phpunit/php-token-stream": "^3.1.1",
+        "phpunit/php-token-stream": "^3.1.1 || ^4.0",
         "phpunit/php-text-template": "^1.2.1",
         "sebastian/code-unit-reverse-lookup": "^1.0.1",
         "sebastian/environment": "^4.2.2",


sebastianbergmann:
PHPUnit 8.5 is supported on PHP 8 but without code coverage functionality. See https://github.com/sebastianbergmann/php-code-coverage/blob/7.0/ChangeLog.md#changed and https://github.com/sebastianbergmann/phpunit/blob/8.5/ChangeLog-8.5.md#changed.

That being said, you should not run into an error. Can you please provide instructions on how to reproduce it? PHPUnit 8.5 should not call into php-code-coverage and php-token-stream from there.

sebastianbergmann:
Also note that even at version 4, php-token-stream is not compatible with PHP 8 -- and will not be made compatible with PHP 8.

leofeyer:
You are right, the issue is not directly caused by PHPUnit 8.5. We are using phpunit-token-stream in Contao 4.9 LTS (see [PhpFileLoader@4.9](https://github.com/contao/contao/blob/4.9/core-bundle/src/Config/Loader/PhpFileLoader.php#L50)) and we need to install version 4.0.4. However, PHPUnit 8.5 only allows version 3:

```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - phpunit/phpunit[8.5.11, ..., 8.5.13] require phpunit/php-code-coverage ^7.0.12 -> satisfiable by phpunit/php-code-coverage[7.0.12, 7.0.13].
    - phpunit/phpunit 8.5.10 requires phpunit/php-code-coverage ^7.0.11 -> satisfiable by phpunit/php-code-coverage[7.0.11, 7.0.12, 7.0.13].
    - phpunit/phpunit 8.5.9 requires phpunit/php-code-coverage ^7.0.10 -> satisfiable by phpunit/php-code-coverage[7.0.10, 7.0.11, 7.0.12, 7.0.13].
    - phpunit/phpunit[8.5.0, ..., 8.5.8] require phpunit/php-code-coverage ^7.0.7 -> satisfiable by phpunit/php-code-coverage[7.0.7, ..., 7.0.13].
    - phpunit/php-code-coverage 7.0.7 requires phpunit/php-token-stream ^3.1.0 -> found phpunit/php-token-stream[3.1.0, 3.1.1, 3.1.2] but it conflicts with your root composer.json require (^4.0).
    - phpunit/php-code-coverage[7.0.8, ..., 7.0.13] require phpunit/php-token-stream ^3.1.1 -> found phpunit/php-token-stream[3.1.1, 3.1.2] but it conflicts with your root composer.json require (^4.0).
    - Root composer.json requires phpunit/phpunit ^8.5 -> satisfiable by phpunit/phpunit[8.5.0, ..., 8.5.13].
```

With this PR, it would be possible to require `phpunit/phpunit: ^8.5` and `phpunit/php-token-stream: ^4.0.4`, which is what we would need. 🙈

sebastianbergmann:
Again: php-token-stream is not compatible with PHP 8.

ausi:
But it was marked compatible since version 4.0.3, see https://github.com/sebastianbergmann/php-token-stream/commit/4201f2134700684744e6676392f09a387df4f7ad

leofeyer:
I know. But Contao 4.9 is a long term support version, which will never use any PHP 8-specific code. Therefore it works quite well in our case. 🤷🏻‍♂️

The most recent Contao version uses `nikic/php-parser` instead of php-token-stream, so this is really only about enabling the combination of `phpunit/phpunit: ^8.5` and `phpunit/php-token-stream: ^4.0.4` for an existing LTS version.

leofeyer:
Thank you. 🙏🏻

Issue #842, opened 2021-01-14
Increased memory usage when generating coverage report


tezhm:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 9.2.5
| PHP version               | 7.4.14
| Driver                    | Xdebug
| Xdebug version (if used)  | 3.0.1
| Installation Method       | Composer
| Usage Method              | PHPUnit
| PHPUnit version (if used) | 9.5.0

When generating the coverage report there seems be a large increase in memory required since updating to version ^9. After debugging it seems that the changes to `SebastianBergmann\CodeCoverage\Report\PHP@process` are to blame - namely the addition of `addcslashes()`.

Source:

```php
<?php declare(strict_types=1);
/*
 * This file is part of phpunit/php-code-coverage.
 *
 * (c) Sebastian Bergmann <sebastian@phpunit.de>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */
namespace SebastianBergmann\CodeCoverage\Report;

use function addcslashes;
use function dirname;
use function file_put_contents;
use function serialize;
use function sprintf;
use SebastianBergmann\CodeCoverage\CodeCoverage;
use SebastianBergmann\CodeCoverage\Directory;
use SebastianBergmann\CodeCoverage\Driver\WriteOperationFailedException;

final class PHP
{
    public function process(CodeCoverage $coverage, ?string $target = null): string
    {
        $buffer = sprintf(
            '<?php
return \unserialize(\'%s\');',
            addcslashes(serialize($coverage), "'")  // High memory usage
        );

        if ($target !== null) {
            Directory::create(dirname($target));

            if (@file_put_contents($target, $buffer) === false) {
                throw new WriteOperationFailedException($target);
            }
        }

        return $buffer;
    }
}
```

Before I used to require ~300MB of memory to generate a report, after the update I now require ~768MB of memory.

To demonstrate this I've printed the peak memory usage for each operation:

```php
    public function process(CodeCoverage $coverage, ?string $target = null): string
    {
        echo memory_get_peak_usage() . PHP_EOL;

        $serialize = serialize($coverage);
        echo "Serialize: " . strlen($serialize) . PHP_EOL;

        echo memory_get_peak_usage() . PHP_EOL;

        $cslashes = addcslashes($serialize, "'");
        echo "Cslashes: " . strlen($cslashes) . PHP_EOL;

        echo memory_get_peak_usage() . PHP_EOL;

[...]

PR #844, opened 2021-01-20
Fixing memory usage issues when generating report (#842)

tezhm:
Batching serialized `CodeCoverage` so that less memory is required when generating escaped string via `addcslashes()` method.

### DIFF
--- a/src/Report/PHP.php
+++ b/src/Report/PHP.php
@@ -12,20 +12,25 @@ namespace SebastianBergmann\CodeCoverage\Report;
 use function addcslashes;
 use function dirname;
 use function file_put_contents;
+use function min;
 use function serialize;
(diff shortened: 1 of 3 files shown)


tezhm:
FYI: tried writing to output buffer (ob_start, ob_get_clean) but seemed to use slightly more memory than appending to string variable.

I think a better solution would be to defer the `addcslashes` call to the output file instead and write the `serialize($coverage)` to a temp file as a binary string. The temp file would then be loaded by the output file in a different instance and hopefully have more free memory to use.

dvdoug:
I have an alternate approach for this in #845 that removes the need to use `addcslashes` at all

tezhm:
Awesome. thankyou.

PR #845, opened 2021-01-21
Use a nowdoc instead of escaping the serialized string. Fixes #842

### LINKED ISSUE #842: Increased memory usage when generating coverage report

dvdoug:
Hello @sebastianbergmann 

This PR should fix #842 by using a nowdoc instead of a quote-delimited string meaning that the whole problem of escaping quotes disappears entirely.

Includes a test for the `eval()'d code` case that required the escaping in the first place to avoid any future regression.

The build is failing because of a Composer issue, I cannot see how that might be related to the changes in the PR

### DIFF
--- a/src/Report/PHP.php
+++ b/src/Report/PHP.php
@@ -9,7 +9,6 @@
  */
 namespace SebastianBergmann\CodeCoverage\Report;
 
-use function addcslashes;
 use function dirname;
 use function file_put_contents;
 use function serialize;
@@ -23,9 +22,12 @@ final class PHP
     public function process(CodeCoverage $coverage, ?string $target = null): string
     {
         $buffer = sprintf(
-            '<?php
-return \unserialize(\'%s\');',
-            addcslashes(serialize($coverage), "'")
+            "<?php
+return \unserialize(<<<'END_OF_COVERAGE_SERIALIZATION'%s%s%sEND_OF_COVERAGE_SERIALIZATION%s);",
+            PHP_EOL,
+            serialize($coverage),
+            PHP_EOL,
+            PHP_EOL
         );
 
         if ($target !== null) {
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -1688,6 +1688,42 @@ abstract class TestCase extends \PHPUnit\Framework\TestCase
         return $stub;
     }
 
+    protected function getLineCoverageForFileWithEval(): CodeCoverage
+    {
+        $filter = new Filter;
+        $filter->includeFile(TEST_FILES_PATH . 'source_with_eval.php');
+
+        $coverage = new CodeCoverage(
+            $this->setUpXdebugStubForFileWithEval(),
+            $filter
+        );
+
+        $coverage->start('FileWithEval', true);
+        $coverage->stop();
+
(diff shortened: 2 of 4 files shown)


sebastianbergmann:
Merged manually, thanks.

Issue #846, opened 2021-01-28
Method name should not appear in the method signature attribute of Cobertura XML


derekrprice:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 9.2.5
| PHP version               | 7.4
| Driver                    | Xdebug
| Xdebug version (if used)  | 2.9.8
| Installation Method       | Composer
| Usage Method              | PHPUnit
| PHPUnit version (if used) | 9.5.1

I'm not actually sure that this is a problem with the new cobertura.xml support in php-code-coverage from PR #812 or in AzureDevops, but I suspect the former.  At the least, it is probably out of sync with other Cobertura generators.  The name of methods appears in both the "name" and "signature" attributes in the generated XML, like so:

    <method name="__construct" signature="__construct(string $identifier, League\OAuth2\Server\Repositories\RefreshTokenRepositoryInterface $refreshTokenRepository)

This causes the Azure DevOps code coverage plugin to then render every name twice:
![image](https://user-images.githubusercontent.com/5413632/106147913-6da3fa00-6146-11eb-8bf7-14d92700065b.png)

derekrprice:
I reported this as an issue to Azure DevOps too, as I am unsure who is at fault: https://developercommunity.visualstudio.com/content/problem/1323739/code-coverage-task-renders-method-names-twice.html

sebastianbergmann:
CC @smmccabe

smmccabe:
Officially a function signature is the whole thing, but I looked into the Cobertura java source code and it only does the arguments like `string $identifier, League\OAuth2\Server\Repositories\RefreshTokenRepositoryInterface $refreshTokenRepositor` so I guess that is what we should follow.

I'll make up a PR once I get a spare moment.

sebastianbergmann:
Thank you, Shawn.

PR #847, opened 2021-02-06
Fix method name appearing in signature

SpikeHD:
In reference to #846

Parses out the method name so that only what is inside the brackets is kept.


SpikeHD:
CC @smmccabe

sebastianbergmann:
Merged manually, thanks.

PR #855, opened 2021-04-26
fix count() issue

vaisakhshiva:


sebastianbergmann:
What is this supposed to fix? Thanks!

PR #865, opened 2021-09-16
Add PHP8.1 to CI + associated changes

dvdoug:
Hello @sebastianbergmann 

This PR adds support for PHP 8.1. The core code works absolutely fine, but a few things did need changing
- the current minimum version of `nikic/php-parser` does not work on 8.1, breaking library CI that runs tests with `--prefer-lowest`
- PHP 8.1 tokenises `&` differently from <=PHP 8, which leads to broken test expectations. For the affected ones, I've split the expectations up into a set for PHP < 8.1 and a set for PHP 8.1+. Fortunately, this only affects 3 tests.
- Because I needed to add `lowest` to the CI to test the minimum `nikic/php-parser` version, I took the opportunity to completely sync the matrix and relevant build steps with PHPUnit, so this PR adds Windows builds and JIT builds too.

Please let me know if you want any changes


sebastianbergmann:
Thanks! I have been wanting to make these changes for weeks, if not months, now, but did not have the time to do so. I do not think, though, that with/without JIT is required.

Issue #735, opened 2020-03-05
Feature request: LCOV report


teohhanhui:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | N/A
| PHP version               | N/A
| Driver                    | N/A
| Xdebug version (if used)  | N/A
| Installation Method       | N/A
| Usage Method              | N/A
| PHPUnit version (if used) | N/A

For example, it'd be really useful for https://github.com/coverallsapp/github-action which only supports an LCOV file.

sebastianbergmann:
I do not need this and will not work on this. I will, of course, consider a pull request that implements this.

sebastianbergmann:
https://github.com/php-coveralls/php-coveralls/ does not look abandoned to me, I see regular commits from @jrfnl and @keradus, for instance.

I am inclined to reject (sorry for the strong word) https://github.com/sebastianbergmann/php-code-coverage/pull/869 and https://github.com/sebastianbergmann/phpunit/pull/4787 because it seems that LCOV can be generated from Clover XML.

mlocati:
> https://github.com/php-coveralls/php-coveralls/ does not look abandoned to me

When I wrote that comment, the repo was very silent (no commits for almost an year).

I'll take a look at the updates of it.

jrfnl:
As a frequent user (and occasional contributor) of `php-coveralls`, I can only say that a PHPUnit native format is always preferable as it would eliminate the need for another external dependency, including the need for switching PHP versions during test jobs (needed if the minimum supported PHP version of a project is lower than that of `php-coveralls`).

Having said that, as the format has not been included with PHPUnit so far and the first version in which it _would_ be included would likely be 10.0.0 (?) with a PHP 8.1 minimum, it would still be quite a a while before projects which still support older PHP versions would benefit from a PHPUnit native version.

`php-coveralls` works well enough in the mean time.

sebastianbergmann:
> a PHPUnit native format is always preferable as it would eliminate the need for another external dependency

I understand that. However, I am reluctant to add support for new formats as I may end up having to maintain them.

> would likely be 10.0.0 (?) with a PHP 8.1 minimum

It would definitely be 10.0.0.

jrfnl:
> I am reluctant to add support for new formats as I may end up having to maintain them.

@sebastianbergmann I totally understand. Respect.

sebastianbergmann:
[...]

PR #869, opened 2021-09-29
Feature LCOV code coverage

Bwen:
Feature request issue #735 

Generate an lcov code coverage, which can be tested with `genhtml` which is part of most Debian  derived distribution package `lcov`.
Ref: http://ltp.sourceforge.net/coverage/lcov/geninfo.1.php

Once you have generated the lcov file you can generate the html coverage with the following command:
```
vendor/bin/phpunit --path-coverage --coverage-lcov lcov.info

genhtml lcov.info -o html-coverage-directory --branch-coverage
```


sebastianbergmann:
Please rebase your branch so that it has #870. Thanks!

.gitignore
```diff
@@ -2,6 +2,8 @@
 /.php-cs-fixer.php
 /.php-cs-fixer.cache
 /.phpunit.cache
+/.phpunit.result.cache
```
sebastianbergmann, line 5: This is wrong, please revert.

.gitignore
```diff
@@ -2,6 +2,8 @@
...
 /.psalm/cache
+/psalm.xml
```
sebastianbergmann, line 7: This is wrong, please revert.

Bwen: Done

.gitignore
Bwen: Done

src/Report/Lcov.php
```diff
@@ -0,0 +1,122 @@
...
+{
+    /**
+     * @throws WriteOperationFailedException
+     */
+    public function process(CodeCoverage $coverage, ?string $target = null, ?string $name = null): string
+    {
+        $lcovLines = [];
+        $report    = $coverage->getReport();
+
+        foreach ($report as $file) {
```
dvdoug, line 36: My reading of the LCOV docs suggests that the final `.info` file should ideally not be simply organised by filename like this, but should be organised by test instead - i.e. for each testcase executed there should be a separate `TN:[testname]...end_of_record` block, with each block recording the hit data for the individual test in question.

Without doing that the generated HTML won't (can't) include a per-test breakdown like this one from the docs? http://ltp.sourceforge.net/coverage/lcov/output/example/methods/index-detail.html

src/Report/Lcov.php
```diff
@@ -0,0 +1,122 @@
...
+            $lcovLines[] = 'TN:';
+            $lcovLines[] = 'SF:' . $file->pathAsString();
+            $lcovLines[] = 'FNF:' . $file->numberOfMethods();
+            $lcovLines[] = 'FNH:' . $file->numberOfTestedMethods();
+
```
dvdoug, line 41: When running under Xdebug, I think there should be the raw data available to add `BRDA`/`BRF`/`BRH` segments with branch coverage data? See the text report (https://github.com/sebastianbergmann/php-code-coverage/blob/master/src/Report/Text.php#L74) for determining whether the data is available.

src/Report/Lcov.php
```diff
@@ -0,0 +1,122 @@
...
+            $lines = $this->getLinesFromFile($file);
+
+            foreach ($lines as $lineNumber => $data) {
+                $numberExecution = $data['count'];
+
+                if (isset($data['type']) && $data['type'] == 'method') {
+                    $functionName = $data['name'] ?? 'nameless-' . $lineNumber;
+
+                    $lcovLines[] = "FN:{$lineNumber},{$functionName}";
+                    $lcovLines[] = "FNDA:{$numberExecution},{$functionName}";
```
dvdoug, line 51: Why no `FNF` and `FNH`?

src/Report/Lcov.php
dvdoug: To be clearer, my reading of the `.info` spec suggests something like.

```
foreach ($tests as $test) {
    //TN:
    foreach($files as $file) {
        //SF: 
        // ....
        //end_of_record
    }
}
```

dvdoug: Here's an example I got from the lcov test suite [target.txt](https://github.com/sebastianbergmann/php-code-coverage/files/7270454/target.txt)

Bwen: Thank you for taking the time to review this, I will reorganize it on a per test basis and investigate branches further :+1:

sebastianbergmann: Thank you, @dvdoug, for reviewing this in detail.

Bwen: @dvdoug I tried to wrap my head around refactoring to make it on a test basis, but I feel like `RawCodeCoverageData` is an aggregation of all the tests to source files and thus cannot split the data differently. It would require too much changes...

While I was adding the branch coverage, I noticed that the stats for the functions were off, `--coverage-html` vs `--coverage-lcov` + `genhtml`. But I think it is due to the renders, how they consider if a function is tested or not, one seem to not consider them tested if they are partially tested and the other does... The count for the lines and branches matched.

To note, to have `genhtml` generate the html for the branches you need to add `--branch-coverage`, and of course on the phpunit side needs to add `--path-coverage`.

src/Report/Lcov.php
Bwen: its there at the top of the loop.

src/Report/Lcov.php
Bwen: The data `BRDA`/`BRF/`BRH` has been added to the latest commit. Cheers :+1:

src/Report/Lcov.php
dvdoug: My bad!

src/Report/Lcov.php
dvdoug: >I tried to wrap my head around refactoring to make it on a test basis, but I feel like RawCodeCoverageData is an aggregation of all the tests to source files and thus cannot split the data differently. It would require too much changes...

If you look at the built-in HTML report for a file and hover over a line, there's a tooltip produced that lists the individual test cases affecting that line so the data is available. Although the data is initially organised by file it should be possible to interrogate it to get what you need. On the basis that the order of output into `.info` doesn't matter, something like this?

```
foreach ($report as $file) {
   $tests = $this->getCoverageByTestcase($file);
   foreach ($tests as $test) {
      //TN: [$test]
      //SF: [$file]
      //...
   }
}
```

For the preprocessing, I think something like this should work:
```
private function getCoverageByTestcase($file) {
    $tests = array_keys($report->testData());
    foreach ($tests as $test) {
      $testSpecificData['line'][$test] = array_filter($report->lineCoverageData(), function($test) {...});
      $testSpecificData['branch'][$test] = array_filter($report->functionCoverageData(), function($test) {...});
    }
    return $testSpecificData; 
}

```

Bwen: @dvdoug I will investigate further :detective:

Bwen: @dvdoug Stalling again, I fear it is as I though.

The function `SebastianBergmann\CodeCoverage\Node::lineCoverageData` gives you something like this:
```
string(88) "SebastianBergmann\Environment\RuntimeTest::testGetVersionReturnsPhpVersionWhenRunningPhp"
197 => Array
(
    [0] => SebastianBergmann\Environment\RuntimeTest::testNameAndVersionCanBeRetrieved
    [1] => SebastianBergmann\Environment\RuntimeTest::testNameAndCodeCoverageDriverCanBeRetrieved
    [2] => SebastianBergmann\Environment\RuntimeTest::testGetVersionReturnsPhpVersionWhenRunningPhp
)
```
You have the line number as an index, but the test name will only appear once thus its impossible to tell how many times this specific test case hit that line. Has you can see in the popover in the built-in HTML coverage, it list how many test cases hit the line but **not how many times per test cases**. Which was an indicator to me that the data is just not there.

If organized by test cases in `lcov.info` we cant always put the `total hits` for all test cases lines, it has to be separated into each test cases.
lcov-test-coverage

Unless you can point me to something I missed, I think what is currently there is the best that we can achieve with the data we have.

sebastianbergmann:
Out of curiosity: what are you missing from the HTML report generated by this library that you get from the one generated by LCOV's `genhtml`?

Bwen:
> Out of curiosity: what are you missing from the HTML report generated by this library that you get from the one generated by LCOV's `genhtml`?

To be honest, I prefer the built-in HTML report. I am only using the `genhtml` to test out the `lcov.info` file. The reason for me doing this is purely for convenience for external tools as mentioned in:
https://github.com/sebastianbergmann/php-code-coverage/issues/735

And for that specific case, line coverage is more than enough imo.

src/Report/Lcov.php
dvdoug: >thus its impossible to tell how many times this specific test case hit that line

Are you sure that's what hit means in this context? I read it as 6 out of 8 lines in `iterate.c` were hit (at least once) by `test_2_to_2000`, 7 out of 8 lines were hit (at least once) by `test_overflow` etc. If it were intended to represent just a raw count (e.g. this test case hit the same line 5 times), it wouldn't have the denominator? From the example screenshot the denominators for the individual files (4 + 8) equal the total number of lines (12) displayed at top-right. Same for functions (1 + 1 = 2) and branches (2 + 4 = 6)

tests/_files/BankAccount-lcov-line.info
```diff
@@ -0,0 +1,27 @@
+TN:
+SF:/app/tests/_files/BankAccount.php
```
dvdoug, line 2: `SF:%s/tests/_files/BankAccount.php` should get the CI passing 🤞

Bwen: opps, for to adjust them :facepalm:

src/Report/Lcov.php
Bwen: You are right, wrong example :sweat_smile: . But that is just the rendered doing its thing believe. The fact is we have to provide:
```
DA:<line number>,<execution count>[,<checksum>]
```
**On a test case basis**. And that means we need to know how many time a specific test has hit a line. That is **if** we organize by test cases. If we dont group by test cases (_which is supported in `lcov` format, since it renders_) we wont be getting that extra section for each file. Which isnt a big deal as its meant mostly for external code coverage tools such as coveralls.io which have their github action only type `lcov` format.

tests/_files/BankAccount-lcov-line.info
Bwen: Thats weird, it worked before with  `%s%e`, I was following whats in the `BankAccount-clover-line.xml` :thinking:

src/Report/Lcov.php
dvdoug: > \<execution count\>

That's unfortunate, neither Xdebug or PCOV collect that data so it's unfortunately not a question of changing php-code-coverage to pass it through.

However, I'm afraid I don't get this bit

> If we dont group by test cases (which is supported in lcov format, since it renders) we wont be getting that extra section for each file.

Even if the file were to be organised into one massive unnamed "test" (`TN: <blank>`), you'd still need to include `DA` segments for each because that's the only part that indicates whether an individual line is covered or not. Without it, you'd have no way of telling which lines aren't covered by tests (I note that in fact you are). What you're doing is coalescing the tests together and substituting the number of them that cover a line as the "execution count" as being the "best-available data".

I'm of the opinion that the best-fit to the format is to break it out by test, using a simple 0 or 1 for execution count if that's the best that can be done. For me, that keeps more granularity. It's always possible for LCOV-ingesting tools to combine the data for consumers who don't care about per-test breakdowns, but it's not possible the other way. For the purpose of producing summary statistics, my per-test 0/1 and your <count of tests> would in fact produce the same numbers.

One for @sebastianbergmann to decide!

sebastianbergmann: > I'm of the opinion that the best-fit to the format is to break it out by test, using a simple 0 or 1 for execution count if that's the best that can be done.

Makes sense to me.

Bwen: I'll re-organize it on a test basis and use `0, 1` for line hits :+1: 
Cheers! :partying_face:

src/Report/Lcov.php
```diff
@@ -0,0 +1,154 @@
...
+            $testSpecificData[$test] = [
+                'lineNumbers'            => [],
+                'lineNumbersHit'         => [],
+                'functionLineNumbers'    => [],
+                'functionLineNumbersHit' => [],
+                'branchLineNumbers'      => [],
+            ];
+
+            foreach ($file->lineCoverageData() as $lineNumber => $data) {
+                $testSpecificData[$test]['lineNumbers'][] = $lineNumber;
```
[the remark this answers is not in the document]
Bwen: You were right about this one, was using the wrong check to filter the non-executable lines.

src/Report/Lcov.php
```diff
@@ -0,0 +1,161 @@
...
+                    $lcovLines[]                         = "BRDA:{$lineNumber},0,{$branchId},{$branchHit}";
+                }
+                $lcovLines[] = 'BRF:' . $file->numberOfExecutableBranches();
+                $lcovLines[] = 'BRH:' . $file->numberOfExecutedBranches();
+
+                $lcovLines[] = 'end_of_record';
+            }
+        }
+
+        $buffer = implode("\n", $lcovLines) . "\n";
```
dvdoug, line 68: I understand this is how the other reports do it (generate in memory and then write to disk), but I'm concerned these might turn out to be very, very large files.

What's the largest project you've tested this with?

Bwen: > `fwrite` is probably a better choice? `file_put_contents` would open and close the file on every call.
> 

What do I do about returning the "buffer" at the end? Just open up the file and return the whole thing as string? :thinking:

tests/tests/LcovTest.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        $lcov = new Lcov;
+
+        $this->assertStringMatchesFormatFile(
+            TEST_FILES_PATH . 'BankAccount-lcov-path.info',
+            $lcov->process($this->getPathCoverageForBankAccount(), null, 'BankAccount')
+        );
+    }
+
+    public function testCloverForFileWithIgnoredLines(): void
```
IonBazan, line 39:
```suggestion
    public function testLcovForFileWithIgnoredLines(): void
```

tests/tests/LcovTest.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        $lcov = new Lcov;
+
+        $this->assertStringMatchesFormatFile(
+            TEST_FILES_PATH . 'ignored-lines-lcov.info',
+            $lcov->process($this->getCoverageForFileWithIgnoredLines())
+        );
+    }
+
+    public function testCloverForClassWithAnonymousFunction(): void
```
IonBazan, line 49:
```suggestion
    public function testLcovForClassWithAnonymousFunction(): void
```

src/Report/Lcov.php
dvdoug: > What do I do about returning the "buffer" at the end? Just open up the file and return the whole thing as string?

It's only the text report where the contents are reused by the calling application 🤔. @sebastianbergmann would you be OK if nothing was returned?

Bwen: its used for the tests as well, returning an empty string would make the lcov tests fail... I would have to refactor to test it another way... :thinking:

dvdoug:
I will defer to @sebastianbergmann on adding options, I'm just a humble contributor.

sebastianbergmann:

PR #890, opened 2022-02-07
Remove useless code

Slamdunk:
See https://github.com/sebastianbergmann/php-code-coverage/pull/868#discussion_r800496605

The method does nothing: `unset($file);` has no effect on anything, it's just an unused local array key


dvdoug:
Apparently I had a major brain fart when writing that, and I should also look at why the test didn't reproduce the scenario accurately enough. However, the issue it was added for (#864) is real so I'd prefer fixing this up to do what it should be doing (`unset($this->lineCoverage[$file], $this->functionCoverage[$file])`) rather than removing it.

Issue #889, opened 2022-02-04
CodeCoverage depends on autoload order


Slamdunk:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 9.2.10
| PHP version               | 8.1.2
| Driver                    | Tested on both PCOV & Xdebug
| Xdebug version  | 3.1.2
| Installation Method       | Composer
| Usage Method              | PHPUnit
| PHPUnit version | 9.5.13

Hi, currently CodeCoverage depends on code execution order.
I created a self contained repo that shows this behavior in https://github.com/Slamdunk/phpunit_codecoverage_bug#readme

Basically, given the following code:

```php
/* 1 */ class A
/* 2 */ {
/* 3 */     public function inArray(): bool
/* 4 */     {
/* 5 */         return \in_array(1, [
/* 6 */             AType::OPT,
/* 7 */         ], true);
/* 8 */     }
/* 9 */ }
```

1. lines 6 and 7 are **not** considered `LOC` if `AType` was already loaded and parsed before the `A` is used in the test
2. otherwise if `AType` is unknown to PHP during `inArray` call, lines 6 and 7 **are** considered `LOC`.

This messes CodeCoverage merging in context like ParaTest where test execution is isolated but not deterministic.

Merging two CC reports coming from different tests run in different processes can result in something like this:

![image](https://user-images.githubusercontent.com/152236/152554933-1380932c-f27a-4e77-99c3-d473cbcb620d.png)

Which doesn't make much sense

I don't know if this library can do much about this, maybe @krakjoe and @derickr could help

dvdoug:
Can you try changing `processUncoveredFiles="true"` to `processUncoveredFiles="false"` and see if that helps?

dvdoug:
Noting that it's class constants involved here which form part of to class definition , I'm thinking for Xdebug, this relates to/is another example of https://bugs.xdebug.org/view.php?id=1798 which affects `extends` and `implements`

sebastianbergmann:
CC @krakjoe @derickr

Slamdunk:
> Can you try changing `processUncoveredFiles="true"` to `processUncoveredFiles="false"` and see if that helps?

This has no effect on the underlying issue

> Noting that it's class constants involved here which form part of the class definition, I'm thinking that for Xdebug this might relate to/be another example of https://bugs.xdebug.org/view.php?id=1798 which affects `extends` and `implements`?

[...]

PR #891, opened 2022-02-08
Line coverage stubs: be consistent with actual xDebug/PCOV output

Slamdunk:
Found this inconsistency while investigating https://github.com/sebastianbergmann/php-code-coverage/issues/889

I manually checked both `Xdebug3Driver` and `PcovDriver` outputs and they differ a bit from what's coded in `TestCase`.

Most of the differences are negligible and have no impact on the purpose of the involved tests, they all still pass, but the diff addressed in this PR can sneak bugs in the future if not merged, especially while merging multiple CC reports.

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -50,30 +50,46 @@ abstract class TestCase extends \PHPUnit\Framework\TestCase
             RawCodeCoverageData::fromXdebugWithoutPathCoverage([
                 TEST_FILES_PATH . 'BankAccount.php' => [
                     8  => 1,
+                    9  => -2,
                     13 => 1,
+                    14 => -1,
+                    15 => -1,
                     16 => 1,
+                    18 => -1,
                     29 => 1,
+                    31 => -1,
+                    32 => -2,
                 ],
             ]),
             RawCodeCoverageData::fromXdebugWithoutPathCoverage([
                 TEST_FILES_PATH . 'BankAccount.php' => [
                     8  => 1,
+                    9  => -2,
                     13 => 1,
+                    14 => -1,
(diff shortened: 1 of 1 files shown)


tests/TestCase.php
```diff
@@ -50,30 +50,46 @@ protected function getLineCoverageXdebugDataForBankAccount()
             RawCodeCoverageData::fromXdebugWithoutPathCoverage([
                 TEST_FILES_PATH . 'BankAccount.php' => [
                     8  => 1,
+                    9  => -2,
                     13 => 1,
+                    14 => -1,
+                    15 => -1,
```
Slamdunk, line 56: Reference:

https://github.com/sebastianbergmann/php-code-coverage/blob/caeda52223e37eaddc3ae61e7d4401f4e6821e71/tests/_files/BankAccount.php#L11-L18

### Line 14

Line 14 is **always** reported by each driver. Not having it here can introduce a bug on CC merging in the future

### Line 15

This line is not reported by any driver, but I kept it here because removing it would fail pretty much every test in this library.
I've checked each of them, and removing this reported line to be consistent with drivers output adds no value to the test suite.

Slamdunk:
To be more clear on how I got this PR committed:

1. I've run each of the 4 `BankAccountTest` tests
2. Gathered `xdebug_get_code_coverage()` output
3. Copy-pasted it into `\SebastianBergmann\CodeCoverage\TestCase::getLineCoverageXdebugDataForBankAccount` (and partially into `getPathCoverageXdebugDataForBankAccount`)
4. Manually re-added line `15 => -1,` to the arrays to get the test pass again

Issue #850, opened 2021-02-27
Make static analysis cache usable during CI


sanmai:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 7.0.14 

It isn't new that getting a coverage report is kind of slow in recent versions of the package, as explored in #789 in great detail. Yet we can't get any benefits from the cache during CI, looking at about 40-50% elevated times. For a medium-sized project it could be additional 5 minutes for CI to pass, even more for larger projects.

### The root of the issue

Static analysis cache uses modification time to validate cache, yet during CI we have a fresh checkout which has all timestamps reset to current time. That's what Git does. Which event, in turn, invalidates all cache no matter what effort we make to restore it.

https://github.com/sebastianbergmann/php-code-coverage/blob/0b4ce85a86040303d93beddf4f31f2ab07775e9a/src/StaticAnalysis/Cache.php#L36-L49

On the other hand, we truly shouldn't be using stale caches, so we can't just reset all `mtimes` in cache.

### Possible solution

Static analysis cache storage can use a hash of a whole file as a part of a final hash, thus ensuring caches are busted if a file was modified. This can be done by default, or on an opt-in basis, keeping old `mtime` logic intact. 

(I don't think if we'd read a file twice would matter much because chances are it'll be a cached read. So it might be just as good to introduce a new behavior forgetting the old.)

sanmai:
Please let me know if this is a good problem to work on.

sebastianbergmann:
@Slamdunk @dvdoug What do you think? Should we use a hash of a file's contents instead of its `filemtime()`?

sanmai:
Fixed in db1c691dac2b21f449b4c21d219be090c5adcc12

dvdoug:
> @Slamdunk @dvdoug What do you think? Should we use a hash of a file's contents instead of its `filemtime()`?

Seems like a good idea to me

PR #898, opened 2022-02-21
Static analysis cache: rely on content hash

Slamdunk:
Fix https://github.com/sebastianbergmann/php-code-coverage/issues/850

### DIFF
--- a/src/StaticAnalysis/ParsingFileAnalyser.php
+++ b/src/StaticAnalysis/ParsingFileAnalyser.php
@@ -34,6 +34,11 @@ final class ParsingFileAnalyser implements FileAnalyser
     /**
      * @var array
      */
+    private $hash = [];
+
+    /**
+     * @var array
+     */
     private $classes = [];
 
     /**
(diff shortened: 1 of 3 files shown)


src/StaticAnalysis/ParsingFileAnalyser.php
```diff
@@ -71,12 +76,24 @@ final class ParsingFileAnalyser implements FileAnalyser
      */
     private $ignoreDeprecatedCode;
 
+    public static function computeHashForSource(string $source): int
+    {
+        return crc32($source);
```
Slamdunk, line 81: 1. A single method to not duplicate code
2. PHP-CS-Fixer uses `crc32` without known issues: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/ae5701a7d22f3a808accf6e63e7b61abf8099ce6/src/Cache/FileCacheManager.php#L127

sebastianbergmann:
Thanks!

I have changed your implementation in db1c691dac2b21f449b4c21d219be090c5adcc12 and made the hash part of the filename.

Issue #556, opened 2017-11-09
Make colors in HTML report configurable


sebastienbarre:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 4.0.8
| PHP version               | 7.0.24
| Driver                    | Xdebug
| Xdebug version (if used)  | 2.5.5
| Installation Method       | Composer
| Usage Method              | PHPUnit
| PHPUnit version (if used) | 5.7.23

Hello.
Is there an easy way to customize the colors in [style.css](https://github.com/sebastianbergmann/php-code-coverage/blob/master/src/Report/Html/Renderer/Template/css/style.css)? My variant of colorblindness make it pretty difficult to differentiate between high coverage (`#dff0d8`) and low coverage (`#f2dede`) as they both have very low saturation (8% to 10%). Thank you.

sebastianbergmann:
The easiest way to do this would be to have a custom `style.css` that you copy to the target directory, for instance in the test target of your build automation.

sebastienbarre:
I mostly do coverage manually, e.g. `,.Vendor/bin/phpunit --coverage-html coverage Test/Case`. Is there a configurable hook there where I could sneak a `cp` or something?

sebastianbergmann:
I will look into making the colors configurable.

sebastienbarre:
Thank you, @sebastianbergmann

sebastianbergmann:
I just realized that my knowledge about CSS in general and Bootstrap in particular is insufficient to figure out the "right way" to change the colors of all elements affected by `success`, `warning`, and `danger` in `style.css`. In order to implement this I need someone to tell me how to do this.

kelunik:
@sebastianbergmann Maybe we can just have better defaults instead of making it configurable? How about going with black (low) → green (high), @sebastienbarre does that work?

matthewtrask:
I think darker reds and greens would be good? It still holds the same sumbolization but easier to see.

kelunik:
@matthewtrask Changing the saturation / shade of both might lead to the same issues again for color blind people.

sebastianbergmann:
I would like to keep the current colors as default.

sebastianbergmann:
Of course. All I need is an answer to https://github.com/sebastianbergmann/php-code-coverage/issues/556#issuecomment-344356203.

sebastienbarre:
[...]

PR #901, opened 2022-02-21
Add interface to specify custom.css file for HTML report

hemberger:
Related to #556.

It would help to make the custom CSS feature more accessible by making
the file a configuration option, e.g. in phpunit:

    <coverage>
        <report>
            <html customCssFile="path/to/custom.css" />
        </report>
    </coverage>

The `$customCssFile` argument is added to the Html\Facade interface as
a prerequisite to any changes to phpunit configuration options. If not
null, it copies this file to the custom.css destination instead of the
empty custom.css stub file.

-----------------------

A couple logistical notes:
1. There wasn't an obvious way to add a good test for this feature. If there's something specific you'd like to see tested, I'm more than happy to add it.
2. If you're on board with this plan, I can also put together a PR in phpunit to add the configuration option. Is there any protocol for adding features to phpunit that depend on a specific version of php-code-coverage?

Thanks for your time!

### DIFF
--- a/src/Report/Html/Facade.php
+++ b/src/Report/Html/Facade.php
@@ -41,7 +41,12 @@ final class Facade
      */
     private $highLowerBound;
 
-    public function __construct(int $lowUpperBound = 50, int $highLowerBound = 90, string $generator = '')
+    /**
+     * @var ?string
+     */
+    private $customCssFile;
+
+    public function __construct(int $lowUpperBound = 50, int $highLowerBound = 90, string $generator = '', ?string $customCssFile = null)
     {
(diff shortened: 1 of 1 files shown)


sebastianbergmann:

I think that this should be handled outside of php-code-coverage (and PHPUnit). Your build automation should copy your `custom.css` to the `css` directory in the generated directory.

sebastianbergmann:
Meditate on this, I shall.

sebastianbergmann:
Superseded by cd4b129b5194b6ffac46f7022b925019b603231a.

Issue #905, opened 2022-02-26
Regression in empty constructor coverage


hemberger:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 9.2.11+
| PHP version               | 8.1.3
| Driver                    | Xdebug
| Xdebug version (if used)  | 3.1.3
| Installation Method       | Composer
| Usage Method              | PHPUnit
| PHPUnit version (if used) | 9.5.14

Coverage of empty constructors has changed between versions 9.2.10 and 9.2.11, and now they are no longer coverable code. In some sense this is not unreasonable, because they are empty after all, but I'm not sure if this was an intentional change. I kind of prefer that they be coverable, since it is still code that can be executed.

Here is an example class:
```php
<?php declare(strict_types=1);

namespace Smr;

class Bla {

    public function __construct(
        public string $foo,
    ) {}

}
```

And the test that covers it:
```php
<?php declare(strict_types=1);

use Smr\Bla;

/**
 * @covers Smr\Bla
 */
class BlaTest extends \PHPUnit\Framework\TestCase {

    public function test_ctor() : void {
        $bla = new Bla('a');
        self::assertSame('a', $bla->foo);
    }   

}
```
Here is the coverage report in 9.2.10:
![image](https://user-images.githubusercontent.com/846186/155831967-b9de01d7-2768-4e8d-af80-063204c2924a.png)

Here is the coverage report in 9.2.11 (note that we get the same report in 9.2.13):
![image](https://user-images.githubusercontent.com/846186/155831981-6671bec4-4db8-49de-8781-936d8e20300a.png)

Thanks again for your time!

sebastianbergmann:
CC @Slamdunk

Issue #904, opened 2022-02-26
Regression in match function coverage


hemberger:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 9.2.13
| PHP version               | 8.1.3
| Driver                    | Xdebug
| Xdebug version (if used)  | 3.1.3
| Installation Method       | Composer
| Usage Method              | PHPUnit
| PHPUnit version (if used) | 9.5.14

As recently as php-code-coverage 9.2.11, the builtin `match` function reported coverage for each matching case. In 9.2.13, the cases are not even marked as coverable. Is this expected behavior? It seems quite useful to me to know which `match` cases are covered by tests.

Here is an example class:

```php
<?php declare(strict_types=1);

namespace Smr;

class Bla {

    public int $foo;

    public function __construct(string $bar) {
        $this->foo = match ($bar) {
            'a' => 1,
            'b' => 2,
            'c' => 3,
            // Some comments
            default => 4,
        };
    }   

}
```

And here is the corresponding test:

```php
<?php declare(strict_types=1);

use Smr\Bla;

/**
 * @covers Smr\Bla
 */
class BlaTest extends \PHPUnit\Framework\TestCase {

    public function test_a() : void {
        $bla = new Bla('a');
        self::assertSame(1, $bla->foo);
    }   

    public function test_b() : void {
        $bla = new Bla('b');
        self::assertSame(2, $bla->foo);
    }   

    public function test_default() : void {
        $bla = new Bla('xyz');
        self::assertSame(4, $bla->foo);
    }   

}
```

Here is the coverage report from 9.2.11:
![image](https://user-images.githubusercontent.com/846186/155830866-c92e97b6-f5a7-48e6-9422-2e6fc8b7d93d.png)
(Note that the tests covering cases `'a'`, `'b'`, and `default` are `test_a`, `test_b`, and `test_default` respectively, as expected.)

And here is the coverage report from 9.2.13:
![image](https://user-images.githubusercontent.com/846186/155830911-2c9b0ab4-1550-4e76-90f8-dbafc06e8bc4.png)

Thanks for your time!

sebastianbergmann:
CC @Slamdunk

PR #906, opened 2022-02-28
Executable lines: fix empty constructor and match expression regressions

Slamdunk:
Fix https://github.com/sebastianbergmann/php-code-coverage/issues/904
Fix https://github.com/sebastianbergmann/php-code-coverage/issues/905


hemberger:
Thanks so much for the quick fix!

Issue #907, opened 2022-03-03
Line with only "return [" is not recognized as executable


sebastianbergmann:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 9.2.14
| PHP version               | 8.1.3
| Driver                    | PCOV
| PCOV version (if used)  | 1.0.11
| Installation Method       | Composer
| Usage Method              | PHPUnit
| PHPUnit version (if used) | 9.5.16

![grafik](https://user-images.githubusercontent.com/25218/156604560-1584438e-e366-4470-9009-f80b790ac8eb.png)

#### `Example.php`
```php
<?php declare(strict_types=1);

namespace example;

final class Example
{
    public function one(): array
    {
        return [
            '...',
        ];
    }

    public function two(): array
    {
        return ['...',
        ];
    }


    public function three(): array
    {
        return ['...',];
    }
}
```
#### `ExampleTest.php`
```php
<?php declare(strict_types=1);

namespace example;

use PHPUnit\Framework\TestCase;

/**
 * @covers \example\Example
 */
final class ExampleTest extends TestCase
{
    public function testOne(): void
    {
        $this->assertSame(['...'], (new Example)->one());
    }
}
```

Slamdunk:
Yes I have: xDebug & PCOV consider `'...'` the LoC, while our visitor only watch for the `return` statement.

I can fix this, but not within days

sebastianbergmann:
Thanks! Take your time.

Slamdunk:
This is what xDebug reports:

```php
<?php

class Foo
{
    public function one(): array
    {
        return [
            '...',                // <---
            '...',
        ];
    }

    public function two(): array
    {
        return ['...',            // <---
            '...',
        ];
    }

    public function three(): array
    {
        return ['...',];          // <---
    }

    public function four(): bool
    {
        return in_array('...', [  // <---
            '...',
            '...',
        ], true);
    }
}
```

The fourth example changes its predictability. Let's see if I can get everything working

PR #909, opened 2022-03-07
Return statement: evaluate content for correct LoC markings

Slamdunk:
Fix https://github.com/sebastianbergmann/php-code-coverage/issues/907

### DIFF
--- a/tests/tests/Data/RawCodeCoverageDataTest.php
+++ b/tests/tests/Data/RawCodeCoverageDataTest.php
@@ -272,6 +272,7 @@ final class RawCodeCoverageDataTest extends TestCase
             [
                 12,
                 14,
+                15,
                 16,
                 18,
             ],
@@ -299,6 +300,7 @@ final class RawCodeCoverageDataTest extends TestCase
             [
                 7,
                 9,
+                10,
                 11,
                 13,
             ],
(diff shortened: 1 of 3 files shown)


tests/tests/Data/RawCodeCoverageDataTest.php
```diff
@@ -362,6 +364,7 @@ public function testHeavyIndentationIsHandledCorrectly(): void
                 117,
                 120,
                 123,
+                125, // This shouldn't be marked as LoC, but it's high unlikely to happen IRL to have $var = []; on multiple lines
```
Slamdunk, line 367: I've thoroughly checked each test that needed to be changed and everything is an edge case now fixed, except this one.

`$var = [];` on multiple lines will be marked on `[]` line too with this PR, which is wrong, but it's unlikely to happen on real code.

As far as I can tell, it's necessary to introduce this little bug to fix much more common ones.

sebastianbergmann:
You are amazing, @Slamdunk. Thank you!

tests/tests/Data/RawCodeCoverageDataTest.php
sebastianbergmann: Makes sense to me.

PR #927, opened 2022-08-19
Use consistent color in first column of HTML Report

### LINKED ISSUE #926: File view has wrong colouring for the first column
| php-code-coverage version | 9.2.15
| PHP version               | 7.4.24
| Driver                    | PCOV
| PCOV version (if used)    | 7.4.13
| Usage Method              | Codeception

The file coverage report still has the method names coloured the same as what is now the last column: Classes and Traits. It should be coloured the same as the second column: Lines.

image

image

In the example screenshot, `medium_method` should be coloured yellow, just like "Lines" is. This will make "file view" fully consistent with "directory view" where the first column can be any colour, not just red or green. Note that "Total" and the file name / class name lines are currently inconsistent too, being yellow in the directory view and red in the file view. "Total" can't be both red and yellow at the same time!

P.S. The last column (Classes and Traits) possibly shouldn't have any colour for method lines since methods aren't classes or traits. I can make that a separate issue though.

hemberger:
Fixes #926.

The first column in the HTML Report will now be colored by the "Lines"
coverage level regardless of view (Directory vs. File) or item type
(Directory, File, Class, or Method).

While this is somewhat a matter of preference, I think the following
arguments support this change:

* Prior to #900, the first column was always colored the same as the
  second column. This change restores that pattern since the second
  column is always "Lines" as of #900.

* In the File view, the first column for Class and Method items were
  colored by "Class" and "Method" coverage respectively. This meant
  that they were only ever 0/1 or 1/1 (red or green). By switching to
  "Lines" coverage for both, the coloring can be more granular.

Previous behavior:
image

New behavior:
image

### DIFF
--- a/src/Report/Html/Renderer/Template/file_item.html.dist
+++ b/src/Report/Html/Renderer/Template/file_item.html.dist
@@ -1,5 +1,5 @@
       <tr>
-       <td class="{{classes_level}}">{{name}}</td>
+       <td class="{{lines_level}}">{{name}}</td>
        <td class="{{lines_level}} big">{{lines_bar}}</td>
        <td class="{{lines_level}} small"><div align="right">{{lines_executed_percent}}</div></td>
        <td class="{{lines_level}} small"><div align="right">{{lines_number}}</div></td>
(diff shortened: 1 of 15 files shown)


sebastianbergmann:
Thank you, @hemberger!

PR #928, opened 2022-08-29
Reorder checks in Filter::isExcluded to touch the FS only when needed

mvorisek:
This change reduces `is_file` calls in short coverage requests significantly, as a lot of `is_file` calls of (normally excluded/not-included) vendor files will not be needed.

Also reduces `Filter::$isFileCache` cache size.

### DIFF
--- a/src/Filter.php
+++ b/src/Filter.php
@@ -100,11 +100,7 @@ final class Filter
 
     public function isExcluded(string $filename): bool
     {
-        if (!$this->isFile($filename)) {
-            return true;
-        }
-
-        return !isset($this->files[$filename]);
+        return !isset($this->files[$filename]) || !$this->isFile($filename);
     }
 
     /**


sebastianbergmann:
Your pull request contains changes (`!$filename` -> `$filename === false`, for instance) that are not related to the topic of your pull request.

mvorisek:
> Your pull request contains changes (`!$filename` -> `$filename === false`, for instance) that are not related to the topic of your pull request.

Yes, this PR also improves the checking of `realpath` failure. It that change welcomed?

sebastianbergmann:
> Yes, this PR also improves the checking of `realpath` failure. It that change welcomed?

Likely, but please separate the two changes.

mvorisek:
PR updated, now it contains only the reordered checks.

PR #929, opened 2022-08-29
Fix FS caching with open_basedir

mvorisek:
`GlobIterator(__DIR__ . '/*.php')` is not working /w open_basedir set in PHP 7.4 - 8.1

in upcoming PHP 8.2 it was fixed in https://github.com/php/php-src/pull/9120

we can simply use `SebastianBergmann\FileIterator\Facade` as in `Filter` class, this FS iterator is working /w open_basedir and it very optimized

### DIFF
--- a/src/StaticAnalysis/CachingFileAnalyser.php
+++ b/src/StaticAnalysis/CachingFileAnalyser.php
@@ -9,15 +9,13 @@
  */
 namespace SebastianBergmann\CodeCoverage\StaticAnalysis;
 
-use function assert;
 use function crc32;
 use function file_get_contents;
 use function file_put_contents;
 use function is_file;
 use function serialize;
-use GlobIterator;
 use SebastianBergmann\CodeCoverage\Util\Filesystem;
-use SplFileInfo;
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Just as PHPUnit, this library is intended to be used in development environments only, not in production environments. I therefore see no reason why it should care about `open_basedir` and work around issues related to this or other security-related settings.

sebastianbergmann:
This PR does more than one thing: it replaces the usage of `GlobIterator` with `SebastianBergmann\FileIterator` to address the issue mentioned in the PR title. However, it also changes how cache keys are generated (`crc32(...)` -> `hash('sha256', ...`).

mvorisek:
> Just as PHPUnit, this library is intended to be used in development environments only, not in production environments. I therefore see no reason why it should care about `open_basedir` and work around issues related to this or other security-related settings.

Even PHPUnit can be run with `open_basedir` and we do run it with this configuration, to prevent phpunit wide system modifications. Every other phpunit part supports it 👍



> This PR does more than one thing...it also changes how cache keys are generated (`crc32(...)` -> `hash('sha256', ...`).

Removed from this PR, will send another PR.

PR #930, opened 2022-08-29
Cache static analysis in system temp directory by default

mvorisek:
Starting with https://github.com/sebastianbergmann/php-code-coverage/pull/892 (released in 9.2.10) static analysis is used for every coverage run which implies there is an huge analysis overhead.

Projects that run coverage analysis in multiple separate processes need about 4x more time to analyse since 9.2.10 release.

Same was reported also here: https://github.com/sebastianbergmann/php-code-coverage/pull/892#discussion_r802971069

This PR set system temp directory for caching by default. If user wants, he can disable the caching or change the caching using existing method. But for general usage, hash based cache does not impose any issues and the speedup in consequent runs in significant.

### DIFF
--- a/src/CodeCoverage.php
+++ b/src/CodeCoverage.php
@@ -22,6 +22,7 @@ use function get_class;
 use function is_array;
 use function is_file;
 use function sort;
+use function sys_get_temp_dir;
 use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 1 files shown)


sebastianbergmann:

mvorisek:
Can I please ask you for short why, what potential problem you see with this PR?

PR #931, opened 2022-08-30
Use md5 for cache hashing instead of weak crc32

mvorisek:
currently collision (with p=50%) is expected with `sqrt(2*2^32*0.5)` [1] = `65536` files only

[1] https://en.wikipedia.org/wiki/Birthday_problem#Square_approximation

### DIFF
--- a/src/StaticAnalysis/CachingFileAnalyser.php
+++ b/src/StaticAnalysis/CachingFileAnalyser.php
@@ -9,10 +9,11 @@
  */
 namespace SebastianBergmann\CodeCoverage\StaticAnalysis;
 
-use function crc32;
 use function file_get_contents;
 use function file_put_contents;
+use function implode;
 use function is_file;
+use function md5;
 use function serialize;
 use SebastianBergmann\CodeCoverage\Util\Filesystem;
 use SebastianBergmann\FileIterator\Facade as FileIteratorFacade;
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Looks like you started working on this before af80b7771102a96c677665d7e2ed69f2a79e9393 and/or 2e94d0330fcadc702935571289b4b8940cf0e599 and now there is a conflict. Can you please resolve it? Thanks!

mvorisek:
rebased

src/StaticAnalysis/CachingFileAnalyser.php
```diff
@@ -159,7 +159,8 @@ private function write(string $filename, $data): void
 
     private function cacheFile(string $filename): string
     {
-        return $this->directory . DIRECTORY_SEPARATOR . hash('sha256', $filename . crc32(file_get_contents($filename)) . self::cacheVersion());
+        return $this->directory . DIRECTORY_SEPARATOR
+            . md5($filename . "\0" . file_get_contents($filename) . "\0" . self::cacheVersion()) . '.bin';
```
sebastianbergmann, line 163: Why the `.bin` suffix?

src/StaticAnalysis/CachingFileAnalyser.php
```diff
@@ -168,13 +169,14 @@ private static function cacheVersion(): string
...
+        $buffer = [];
 
         foreach ((new FileIteratorFacade)->getFilesAsArray(__DIR__, '.php') as $file) {
-            $buffer .= file_get_contents($file);
+            $buffer[] = $file;
+            $buffer[] = file_get_contents($file);
         }
 
-        self::$cacheVersion = (string) crc32($buffer);
+        self::$cacheVersion = md5(implode("\0", $buffer));
```
sebastianbergmann, line 179: The `implode()` function is not imported.

mvorisek: fixed

src/StaticAnalysis/CachingFileAnalyser.php
mvorisek: reverted, not needed

PR #934, opened 2022-09-09
Fix cobertura package name

tm1000:
In certain utilities that parse Cobertura files they look to the `<Package name=` attribute. In PHPUnit this attribute is always empty (null)

This causes results to look like the following (in this parser at least):
image

As you can tell this makes no sense (The parser on the other end adds "[filename] Package 1+n" whenever it encounters an empty package name)

See: https://github.com/sebastianbergmann/phpunit/blob/main/src/TextUI/TestRunner.php#L479

The third parameter is never passed.

It appears that the Cobertura report generator was copied from the Clover report generator but they are not one-in-the same. If `$name` was properly set in the Cobertura report then every "package" would have the same name. This would confuse Cobertura parsers because the way PHPUnit works with each package is it's essentially each `$report` therefore we should be setting the packageName the same as the filename (or the class but filename seems more flexible when the file contains no classes)

Summary:
In a Clover file the "name" attribute is global, see: https://github.com/sebastianbergmann/php-code-coverage/blob/main/src/Report/Clover.php#L46

In Cobertura the "name" attribute is attached to PackageName and a package is generated for each report.

This change isn't really breaking because if name would have ever been passed it would have been added to each package and completely confused any parser (worse than it is now)

### DIFF
--- a/src/Report/Cobertura.php
+++ b/src/Report/Cobertura.php
@@ -25,7 +25,7 @@ final class Cobertura
     /**
      * @throws WriteOperationFailedException
      */
-    public function process(CodeCoverage $coverage, ?string $target = null, ?string $name = null): string
+    public function process(CodeCoverage $coverage, ?string $target = null): string
     {
         $time = (string) time();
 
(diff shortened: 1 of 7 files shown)


tm1000:
Closed in favor of #935 which is pointed towards 9.2 instead of main

PR #936, opened 2022-09-15
Use cached Filter::isFile() instead of is_file()

mvorisek:


mvorisek:
@sebastianbergmann any idea how to fix the psalm error:

```
Run ./tools/psalm --config=.psalm/config.xml --no-progress --shepherd --show-info=false --stats

ERROR: UnresolvableInclude - src/CodeCoverage.php:546:17 - Cannot resolve the given expression to
       a file path (see https://psalm.dev/106)
                include_once $uncoveredFile;
```

PR #938, opened 2022-09-15
Fix coverage for uncovered files with arrays

mvorisek:
`PhpParser\Node\Expr\Array_` is not an executable the same as any scalar

introduced in https://github.com/sebastianbergmann/php-code-coverage/pull/909/files

fixes https://github.com/sebastianbergmann/php-code-coverage/issues/942

also fixes https://github.com/sebastianbergmann/php-code-coverage/pull/909/files#r820517185

### DIFF
--- a/tests/tests/Data/RawCodeCoverageDataTest.php
+++ b/tests/tests/Data/RawCodeCoverageDataTest.php
@@ -272,7 +272,6 @@ final class RawCodeCoverageDataTest extends TestCase
             [
                 12,
                 14,
-                15,
                 16,
                 18,
             ],
@@ -300,7 +299,6 @@ final class RawCodeCoverageDataTest extends TestCase
             [
                 7,
                 9,
-                10,
                 11,
                 13,
             ],
@@ -342,7 +340,6 @@ final class RawCodeCoverageDataTest extends TestCase
                 25,
                 28,
                 31,
-                36,
                 40,
                 46,
                 48,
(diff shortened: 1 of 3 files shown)


mvorisek:
@Slamdunk can you please have a look?

sebastianbergmann:
CC @Slamdunk

tests/tests/Data/RawCodeCoverageDataTest.php
```diff
@@ -415,14 +411,15 @@ public function testReturnStatementWithOnlyAnArrayWithScalarReturnsTheFirstEleme
 
         $this->assertEquals(
             [
-                8,
+                7,
```
Slamdunk, line 414: In order for our `\SebastianBergmann\CodeCoverage\StaticAnalysis\ExecutableLinesFindingVisitor` to work properly it **MUST** output a _subset_ of the lines discovered by CC Drivers.

Run this script to get the superset (after fixing `Foo::seventh` return type to `string`):

```
cat <<'EOF' | XDEBUG_MODE=coverage php
<?php
xdebug_start_code_coverage();
require "tests/_files/source_with_return_and_array_with_scalars.php";
$a = new Foo();
foreach (get_class_methods($a) as $method) {
    $a->$method();
}
print_r(xdebug_get_code_coverage());
EOF
```

And you'll get:

```
    [tests/_files/source_with_return_and_array_with_scalars.php] => Array
        (
            [8] => 1
            [15] => 1
            [24] => 1
            [30] => 1
            [40] => 1
            [47] => 1
            [54] => 1
            [63] => 1
            [66] => 1
        )
```

In the current state, this PR would produce an empty final CC report over arrays.

mvorisek:
closing in favor of https://github.com/sebastianbergmann/php-code-coverage/pull/949

Issue #856, opened 2021-04-30
Do not include (and execute) uncovered files


mvorisek:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 9.5.4
| PHP version         | 8.0.5 with Xdebug 3.0.4
| Installation Method | Composer

#### Summary

see https://github.com/atk4/ui/pull/1628/checks?check_run_id=2476898808#step:9:45

phpunit finishes successfully, but when coverage is disabled, it does fail

#### Current behavior

```
Generating code coverage report in PHP format ... done [00:00.033]
Class "Behat\MinkExtension\Context\RawMinkContext" not found
Error: Process completed with exit code 2.
```

#### How to reproduce

see https://github.com/atk4/ui/pull/1628/checks?check_run_id=2476898808#step:9:45

#### Expected behavior

When class is not loaded in any phpunit tests, zero coverage must be generated even if parent class is not loadable and no error must be generated.

mvorisek:
Here is minimal reproducible case:

Let's have file `x.php` in `src/` directory with the following content:

```
<?php

namespace Pkg;

class BehatContext extends \NotInstalled\BehatMinkContext
{
}
```

when coverage reporting is enabled, the phpunit currently fails to generate the coverage (empty coverage should be reported).

sebastianbergmann:
I can reproduce this:
```
.
├── composer.json
├── phpunit.xml
├── src
│   ├── Bar.php
│   └── Foo.php
└── tests
    └── FooTest.php
```

#### `phpunit.xml`
```xml
<?xml version="1.0" encoding="UTF-8"?>
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.5/phpunit.xsd"
         bootstrap="vendor/autoload.php"
         cacheResultFile=".phpunit.cache/test-results"
         executionOrder="depends,defects"
         forceCoversAnnotation="true"
         beStrictAboutCoversAnnotation="true"
         beStrictAboutOutputDuringTests="true"
         beStrictAboutTodoAnnotatedTests="true"
         failOnRisky="true"
         failOnWarning="true"
         colors="true"
         verbose="true">
    <testsuites>
        <testsuite name="default">
            <directory suffix="Test.php">tests</directory>
        </testsuite>
    </testsuites>

    <coverage cacheDirectory=".phpunit.cache/code-coverage"
[...]

PR #939, opened 2022-09-15
Compile uncovered files with opcache

mvorisek:
Uncovered executable lines must match executable lines from real coverage drivers. This was discovered to be not easy in https://github.com/sebastianbergmann/php-code-coverage/pull/938.

This is problematic and it seems `opcache_compile_file` is a solution. It very nicely collects all possible coverage points (with zero coverage of course).

~In next major version, we can even remove the `processUncoveredFiles` option.~ (already removed)

### DIFF
--- a/src/CodeCoverage.php
+++ b/src/CodeCoverage.php
@@ -515,17 +515,37 @@ final class CodeCoverage
             $this->filter->files(),
             $this->data->coveredFiles()
         );
+        foreach ($uncoveredFiles as $k => $uncoveredFile) {
+            if (!$this->filter->isFile($uncoveredFile)) {
+                unset($uncoveredFiles[$k]);
+            }
+        }
+
+        if (extension_loaded('Zend OPcache')
+            && opcache_get_status() !== false
+            && opcache_get_status()['opcache_enabled']
+        ) {
+            $this->driver->start();
+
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
I am reluctant to merge a change such as this into a minor version, even more so since it addresses an issue that no longer exists in `main`.

mvorisek:
> I am reluctant to merge a change such as this into a minor version, even more so since it addresses an issue that no longer exists in `main`.

@sebastianbergmann looking into the 10.0.0 changelog, you mean https://github.com/sebastianbergmann/php-code-coverage/issues/856?

Such PR removes `processUncoveredFiles` (good), but does not solve the problem of different uncovered lines when not compiled (or executed). Related with https://github.com/sebastianbergmann/phpcov/issues/122, we are experiencing about 0.5% coverage difference... Took me long hours to figure this problem out.

mvorisek:
In https://github.com/atk4/ui/pull/1852 this change increased our coverage by almost 1% because of https://github.com/sebastianbergmann/php-code-coverage/issues/942

mvorisek:
closing as this solution requires opcache and not yet compiled file, closing in favor of fully fixed executable lines static analyser, most fixes landed in https://github.com/sebastianbergmann/php-code-coverage/pull/949

Issue #946, opened 2022-10-16
Method with "return with multiline constant expr only" is not generating any coverage lines


mvorisek:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 9.2.17 / latest
| PHP version               | 8.0.24
| Driver                    | Xdebug
| Xdebug version  | 3.2.0RC1
| Installation Method       | Composer
| Usage Method              | PHPUnit
| PHPUnit version | 9.5.25

code:

https://github.com/atk4/data/blob/0c6bce5c2ac53589934b02bc0832a4753c283f15/src/Persistence/Sql/Oracle/PlatformTrait.php#L104-L109

is not generating any coverage line:

```
...
    <file name="/__w/data/data/src/Persistence/Sql/Oracle/PlatformTrait.php">
      <class name="Atk4\Data\Persistence\Sql\Oracle\PlatformTrait" namespace="global">
        <metrics complexity="7" methods="5" coveredmethods="5" conditionals="0" coveredconditionals="0" statements="24" coveredstatements="24" elements="29" coveredelements="29"/>
      </class>
      <line num="12" type="method" name="getVarcharTypeDeclarationSQL" visibility="public" complexity="1" crap="1" count="209"/>
      <line num="14" type="stmt" count="209"/>
      ...
      <line num="74" type="stmt" count="231"/>
      <line num="94" type="stmt" count="231"/>
      <line num="95" type="stmt" count="231"/>
      <line num="96" type="stmt" count="231"/>
      <line num="99" type="stmt" count="231"/>
      <line num="101" type="stmt" count="231"/>
      <metrics loc="121" ncloc="101" classes="1" methods="5" coveredmethods="5" conditionals="0" coveredconditionals="0" statements="24" coveredstatements="24" elements="29" coveredelements="29"/>
    </file>
...
```

![image](https://user-images.githubusercontent.com/2228672/196032561-e4b11815-cb89-4475-a465-b04e251cb594.png)

I would expect at least one coverage line for the `getListDatabasesSQL` method to be able to tell if called or not

mvorisek:
Hi @Slamdunk, do you think this can be fixed?

Slamdunk:
It definitively should

mvorisek:
here is minimalistic code to reproduce:

```
class Cl
{
    public function getSql(): string
    {
        return 'select 1 from' . ' from dual';
    }
}
```

generates a coverage line, but

```
class Cl
{
[...]

PR #948, opened 2022-10-21
Fix: return with multiline constant expr only must contain the last line

Slamdunk:
Fixes https://github.com/sebastianbergmann/php-code-coverage/issues/946

### DIFF
--- a/tests/_files/source_with_multiline_constant_return.php
+++ b/tests/_files/source_with_multiline_constant_return.php
@@ -1,0 +1,247 @@
+<?php
+
+class Foo
+{
+    public function BitwiseAnd(): int
+    {
+        return
+            1
+            &
+            1
+        ;
+    }
+
+    public function BitwiseOr(): int
+    {
+        return
+            1
+            |
+            1
+        ;
+    }
+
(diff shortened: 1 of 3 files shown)


tests/tests/Data/RawCodeCoverageDataTest.php
```diff
@@ -361,7 +361,6 @@ public function testHeavyIndentationIsHandledCorrectly(): void
                 99,
                 101,
                 116,
-                117,
```
Slamdunk, line 364: :confetti_ball: Hurray that this also fixes a harmless but tedious long lasting bug :nail_care:

tests/_files/source_with_multiline_constant_return.php
```diff
@@ -0,0 +1,247 @@
...
+        return
+            2
+            !==
+            1
+        ;
+    }
+
+    public function Plus(): int
+    {
+        return
```
mvorisek, line 187: there should be also a test with complex constatnt expr like:

```
return 1 + 2
    - 3
    * 4
    - 5;
```

PR #949, opened 2022-10-22
Fix executable lines analysis

### LINKED ISSUE #942: ExecutableLinesFindingVisitor executable lines does not match the lines from xdebug coverage
| php-code-coverage version | 9.2.17 (latest)
| PHP version               | 8.0.23 (latest)
| Xdebug version  | 3.2.0alpha3 (latest)
| Usage Method              | PHPUnit & other
| PHPUnit version (if used) | 9.5.24 (latest)

An issue, that took me quite a while to figure out what is going on.

In atk4/ui we have the following coverage setup:

a) collect coverage in phpunit tests, config: https://github.com/atk4/ui/blob/33779a9414b624a8faa7e959883d4ba407099dcd/phpunit.xml.dist#L23-L29
b) collect coverage from web requests, config: https://github.com/atk4/ui/blob/33779a9414b624a8faa7e959883d4ba407099dcd/demos/init-app.php#L21-L24 and https://github.com/atk4/ui/blob/33779a9414b624a8faa7e959883d4ba407099dcd/src/Behat/CoverageUtil.php#L38-L43
c) then merge all the coverage files using phpunit/phpcov in https://github.com/atk4/ui/blob/33779a9414b624a8faa7e959883d4ba407099dcd/.github/workflows/test-unit.yml#L223

Coverage collection is working. But as soon as we include `tests/` dir also in b), the coverage decreases. This is wrong, as all `tests/` files are already included in a) and inclusion of an already included file in another coverage collection cannot decrease the merged coverage.

CI repro: https://github.com/atk4/ui/pull/1852/commits - see 2nd and 3rd commit

https://app.codecov.io/gh/atk4/ui/blob/b762d62434df393aad170a7378d3d45876f93f5a/tests/JsTest.php (2nd commit coverage)
https://app.codecov.io/gh/atk4/ui/blob/99f347b7d04567faefafa3b39105042b728e1649/tests/JsTest.php (3rd commit coverage)

It seems the problem is in https://github.com/sebastianbergmann/php-code-coverage/pull/892 or in https://github.com/sebastianbergmann/phpcov.

### LINKED ISSUE #954: Empty method does not generate any line/method coverage
| php-code-coverage version | 9.2.18 / latest
| PHP version               | 8.0.24
| Xdebug version  | 3.2.0RC1
| PHPUnit version | 9.5.25

code: https://github.com/atk4/ui/blob/02596615ea82131cd903d6095eec7fe53e541aef/src/Form/Control/DropdownCascade.php#L116-L119

is not generating any coverage line:

```
...
    <file name="/__w/ui/ui/src/Form/Control/DropdownCascade.php">
      <class name="Atk4\Ui\Form\Control\DropdownCascade" namespace="global">
        <metrics complexity="14" methods="5" coveredmethods="1" conditionals="0" coveredconditionals="0" statements="34" coveredstatements="19" elements="39" coveredelements="20"/>
      </class>
      ...
      <line num="103" type="method" name="getJsValues" visibility="private" complexity="3" crap="3.58" count="2"/>
      <line num="105" type="stmt" count="2"/>
      <line num="106" type="stmt" count="2"/>
      <line num="107" type="stmt" count="0"/>
      <line num="109" type="stmt" count="0"/>
      <line num="113" type="stmt" count="2"/>
      <line num="121" type="method" name="renderView" visibility="protected" complexity="1" crap="1" count="2"/>
      <line num="124" type="stmt" count="2"/>
      <line num="126" type="stmt" count="2"/>
      <metrics loc="129" ncloc="98" classes="1" methods="5" coveredmethods="1" conditionals="0" coveredconditionals="0" statements="34" coveredstatements="19" elements="39" coveredelements="20"/>
    </file>
...
```

image

I would expect at least one coverage line for the `htmlRenderValue` method to be able to tell if it was called or not

mvorisek:
add more tests for #948 and fix found issues to make executable lines really a maximal subset (edge cases can be non-present, but not the other way around, otherwise #942) of xdebug

fix https://github.com/sebastianbergmann/php-code-coverage/pull/909/files#r820517185

fix #942 (nested array test added and tested againt about 50k real LoC)

fix #954 (merged via https://github.com/mvorisek/php-code-coverage/pull/2)

fix #938 (test added)


mvorisek:
Hi @Slamdunk, I added more tests, can you please adjust the static analyser for them?

data generated by xdebug

l53 is not critical to be added, but l125 vs. l127 means whole method will be uncovered (issue #946)

I will then conduct more testing - what is the definition of executable line, should it always be 1:1 with xdebug?

Slamdunk:
Sorry for late reply: I'll answer your questions and review the PR as soon as I can

Slamdunk:
> what is the definition of executable line, what are the cases when it should differ from xdebug coverage output?

Due to the runtime nature of the language, there is no definition of executable line.
Let's take two examples:

```
 1 function a() {
 2    return
 3        ''
 4        .
 5        ''
 6    ;
 7 }
 8 
 9 function b() {
10     return
11         a()
12         .
13         ''
14     ;
15 }
```
* In function `a()` the only executable line reported by a driver will be `5`
* but in the function `b()` the only executable line reported by a driver will be `11`, so **without** `13`

So how did I shape the `ExecutableLinesFindingVisitor`?
Any obvious executable line like a function call or variable fetch or constant fetch is counted.
When no executable line is found, I try my best to match the driver's behavoir aiming at the *smallest* subset possible.

As far as I can tell, there's no bulletproof algorithm.
In https://github.com/mvorisek/php-code-coverage/pull/1 I fixed most of the sensible cases, but I turned down the L125 vs L127 dispute.

Who is going to write in real life the example below?
```php
return
    true
    or
    false
;
```
Almost any similar case involves a function call or a variable fetch or constant fetch: all these three cases are already correctly handled.

sebastianbergmann:
Can this be closed then?

Slamdunk:
> Can this be closed then?

The opposite: as sool as https://github.com/mvorisek/php-code-coverage/pull/1 gets merged, this PR will be updated as well and the CI should pass, with a fix for other edge cases that @mvorisek found

sebastianbergmann:
Ah, so you sent him a PR that will update this PR when merged. Got it, was confused / not fully awake yet. Thanks!

Slamdunk:
@mvorisek awesome: you can now to this PR into "Ready for review" I guess

src/StaticAnalysis/ExecutableLinesFindingVisitor.php
```diff
@@ -242,6 +247,29 @@ private function getLines(Node $node): array
         return [$node->getStartLine()];
     }
 
+    private function getNodeStartLine(NodeAbstract $node): int
+    {
+        if ($node instanceof Node\Expr\BooleanNot ||
```
mvorisek, line 252: what have these operators in common? can you suggest a short comment to place above the if, so the code can be more understandable?

https://www.php.net/manual/en/language.operators.precedence.php

https://github.com/nikic/PHP-Parser/tree/v4.15.1/lib/PhpParser/Node/Expr

Slamdunk: I have no idea. I put all my attention in crafting the tests and making them as close to reality as possible.
Then I blindly adjust the `src/` to get them passing :wink:
Of course I tried to refactor the code looking for repeating patterns at exactly the two links you've also followed, but I found none :frowning_face:

mvorisek: I hunted more test cases and the next line with `Node\Expr\AssignOp\ShiftLeft` is not even needed.

The `Node\Expr\BooleanNot` condition is fixing one test, but breaking other I newly added (`complexAssociativityNa2`). I fixed it by calling `getNodeStartLine` method instead of `$node->getEndLine()`, but other test case `unaryMinusNowdoc` breaks tests again.

Currently the static analyser heavily relies on scalars beiing single line. But this is an incorrect assumption and position of all scalars must be analysed using `getNodeStartLine` to handle heredoc/nowdoc correctly.

Then no special handling for `Node\Expr\BooleanNot` should be needed, more precisely, all unary operators has to be enumerated, as there is no `UnaryOp` interface - https://github.com/nikic/PHP-Parser/blob/v4.15.1/lib/PhpParser/Node/Expr/BooleanNot.php#L7.

Slamdunk: > I hunted more test cases

I'd not going down the rabbit hole too much: just stick to real cases, we can't handle every edge cases.
Being 99.9% effective on real existing code is much easier and simplier than aimint at 100% correctness with a red build

> also `testHeavyIndentationIsHandledCorrectly` test seems to output lines for assign that are not present in xdebug output

That one was my fist test coming from a real example but expanded by my imagination: it was a mistake becase it's not easy to understand and doesn't represent the reality; I've kept it though because it's still a good regression test.


I suggest to keep the good https://github.com/sebastianbergmann/php-code-coverage/pull/949/commits/9fc658244bd7c049fc65785dcc5de61780c79ec5 but to revert https://github.com/sebastianbergmann/php-code-coverage/pull/949/commits/13bc95351a7ae890b19e04fafc67f17fd698d8e0 and let's go live with the fixes with managed to accomplish

sebastianbergmann: > I'd not going down the rabbit hole too much: just stick to real cases, we can't handle every edge cases

I could not agree more. Scalars split across multiple lines are definitely not something we should care about.

mvorisek: simple test case as in `testHeavyIndentationIsHandledCorrectly` like

```
    public function xx(): void
    {
        $v3
            =
            x()
        ;
    }
```

must not emit executable line for assign as no such line is emit with xdebug

I will look into this problem by myself and ask for help if needed.

sebastianbergmann: Nobody should write code like that. And if they do, they're on their own. We should not care about this.

mvorisek: @Slamdunk please review

when resolving any php parser node to line, `getNodeStartLine` method must be always used

kukulich:
Will this PR fix this?

Clipboard01

mvorisek:
@kukulich what lines do you expect covered and why?

Slamdunk:
@kukulich as far as I can tell, your code is coverage sample works as expected

kukulich:
Hmm, it looks I sent bad screenshot because I thought both are caused by the change. So now the right screenshots:

**phpunit/php-code-coverage 9.2.17**

before

**phpunit/php-code-coverage 9.2.18**

after

The problem is that `infection/infection` now think that the line is not covered and reports mutants.

With phpunit/php-code-coverage 9.2.17: `10 mutants were not covered by tests`
With phpunit/php-code-coverage 9.2.18: `68 mutants were not covered by tests`

```diff
29) /home/runner/work/BetterReflection/BetterReflection/src/Reflection/StringCast/ReflectionAttributeStringCast.php:27    [M] Concat

--- Original
+++ New
@@ @@
     {
         $arguments = $attributeReflection->getArguments();
         $argumentsFormat = $arguments !== [] ? " {\n  - Arguments [%d] {%s\n  }\n}" : '';
-        return sprintf('Attribute [ %s ]' . $argumentsFormat . "\n", $attributeReflection->getName(), count($arguments), self::argumentsToString($arguments));
+        return sprintf($argumentsFormat . 'Attribute [ %s ]' . "\n", $attributeReflection->getName(), count($arguments), self::argumentsToString($arguments));
     }
```

mvorisek:
some non-executable (compile time) lines are excluded probably by https://github.com/sebastianbergmann/php-code-coverage/pull/892

this PR does not try to add more lines, this PR tries to exclude lines that are no present in xdebug output when formatted with random newlines between php tokens

@Slamdunk can you please review?

@kukulich please describe the problem in separate PR with a test case and description what should be chaged and why

kukulich:
@mvorisek I'm not sure if I don't understand now.

There's https://github.com/sebastianbergmann/php-code-coverage/issues/946 mentioned in the release notes of [9.2.18](https://github.com/sebastianbergmann/php-code-coverage/blob/main/ChangeLog.md#9218---2022-10-27)

This issue was fixed by PR https://github.com/sebastianbergmann/php-code-coverage/pull/948

And this PR draft has in description "Add more tests for https://github.com/sebastianbergmann/php-code-coverage/pull/948 and fix found issues" so it looks to me it is still the same problem and my screenshots are relevant.

However I can open new issue with the problem if it's better for you 🤷‍♂️

mvorisek:
#946 issue opened by me had one motivation - previously `return <scalar / compile time expr>;` was not generating any coverage, thus leaving method completely uncovered, this is not what I see in your screenshot, so yes, please open a separate issue

tests/tests/Data/RawCodeCoverageDataTest.php
```diff
@@ -272,7 +272,6 @@ public function testUseStatementsAreUncovered(): void
             [
                 12,
                 14,
-                15,
```
Slamdunk, line 275: The first element of an array must be counted as executable line

tests/tests/Data/RawCodeCoverageDataTest.php
```diff
@@ -333,7 +331,6 @@ public function testHeavyIndentationIsHandledCorrectly(): void
 
         $this->assertEquals(
             [
-                9,
```
Slamdunk, line 336: Variable assignment must be counted as executable line

tests/tests/Data/RawCodeCoverageDataTest.php
mvorisek: in your initial PR - https://github.com/sebastianbergmann/php-code-coverage/pull/892 - you show in the "after" screenshot exactly the oposite - would you be ok to send another PR into this PR branch?

Slamdunk: You are right: I thought that at least one line of an array declaration should be marked as executable, an in my first PR I set it to be the line of the square bracket.
But that's not true: like you've done here, it's instead correct to count the entire array declaration as non-executable, if it doesn't contain any inside.

Only the variable assignment then is left to be fixed

tests/tests/Data/RawCodeCoverageDataTest.php
mvorisek: the whole assign: https://github.com/sebastianbergmann/php-code-coverage/blob/384d88deeba1a1b9635b045ed244c8c6b866bab7/tests/_files/source_with_heavy_indentation.php#L9-L13

produces only 1 executable/coverage line in xdebug - line 12

line 9 is not present

Slamdunk: In my checks it does:
```php
<?php

class Foo
{
    private $state = 1;

    public function isOne(): void
    {
        $v3
            =
            (bool)
            AType::OPT
        ;
    }
}

class AType
{
    const OPT = 1;
}
```
xDebug report:
```
    [./tests/_files/source_with_heavy_indentation.php] => Array
        (
            [9] => 1
            [12] => 1
            [14] => 1
            [20] => 1
        )
```
As it always did, and mandatory to have an exec line for such conditions.

Slamdunk:
Please don't forse-push until review is complete, as it makes it a lot harder to track changes and review them efficiently

tests/tests/Data/RawCodeCoverageDataTest.php
mvorisek: What is your xdebug/PHP version? Isn't this the exact problem https://github.com/sebastianbergmann/php-code-coverage/issues/889 why you added executable lines, ie. what is your output /wo `AType` defined (and `isOne` never called - this is how I tested it)?

https://github.com/sebastianbergmann/php-code-coverage/pull/949#issuecomment-1301744174 - do you know the `Compare` button after force-push in GH? it nothing has changed, the compare result is empty.

Slamdunk:
```php
<?php

class Foo
{
    private $state = 1;

    public function isOne(): void
    {
        $v3
            =
            (bool)
            1
        ;
    }
}

(new Foo())->isOne();
```
```console
$ XDEBUG_MODE=coverage php -r 'xdebug_start_code_coverage();$file = "tests/_files/source_with_heavy_indentation.php";require $file;print_r(xdebug_get_code_coverage()[$file]);'
Array
(
    [9] => 1
    [12] => 1
    [14] => 1
    [17] => 1
)

$ php -v
PHP 8.1.11 (cli) (built: Sep 29 2022 22:29:14) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.11, Copyright (c) Zend Technologies
    with Zend OPcache v8.1.11, Copyright (c), by Zend Technologies
    with Xdebug v3.1.5, Copyright (c) 2002-2022, by Derick Rethans
```

mvorisek: in
```
PHP 7.4.19 (cli) (built: May  4 2021 14:24:16) ( NTS Visual C++ 2017 x64 )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
    with Zend OPcache v7.4.19, Copyright (c), by Zend Technologies
    with Xdebug v3.0.4, Copyright (c) 2002-2021, by Derick Rethans
```

and script:
```
<?php

$file = realpath(__DIR__ . '/../../phpunit/php-code-coverage/tests/_files/source_with_heavy_indentation.php');

xdebug_start_code_coverage(XDEBUG_CC_UNUSED | XDEBUG_CC_DEAD_CODE);
require $file;
print_r(xdebug_get_code_coverage()[$file]);
```

I get:

```
Array
(
    [12] => -1
    [16] => -1
    [18] => -1
    [19] => -1
    [20] => -1
    [21] => -1
    [22] => -1
    [24] => -1
...
```

sebastianbergmann: Different versions of PHP compile sourcecode to different bytecode. Different bytecode means different code coverage results from Xdebug or PCOV.

Slamdunk: Indeed, line `9` is considered executable line only from PHP 8.0, but in PHP 7.4 it isn't.

What would you like to do about this @sebastianbergmann ?
Different `php-code-coverage` behaviour for different PHP versions, of just push for the most recent version?

sebastianbergmann: I think we should use the behaviour of the current stable PHP version.

mvorisek: @Slamdunk any idea how line 9 can be added? I tried adding `PhpParser\Node\Expr\Variable` node as executable as proposed in https://github.com/sebastianbergmann/php-code-coverage/pull/955 but then the "computeReturns" is broken and line 12 is not reported

see my last commit with a test case, assign can be more complex due array desctructing and both lines 97 and 98 are present in xdebug (tested with PHP 8.1)

mvorisek:
```
        $v2
            =
            (bool)
            1
        ;
```
is very interesting test case

line 1 or line 4 is reported, depending if cached (in opcached) or not, so assign is not stable enought to produce any executable line - it makes sense, no user code is executed

mvorisek: @Slamdunk PR is done, before I squash the changes, please review the changes

Slamdunk:
In the meantime, https://github.com/sebastianbergmann/php-code-coverage/issues/953 gets fixed by https://github.com/mvorisek/php-code-coverage/pull/2

tests/tests/Data/RawCodeCoverageDataTest.php
mvorisek: documented in tests

tests/tests/Data/RawCodeCoverageDataTest.php
```diff
@@ -460,6 +466,38 @@ public function testReturnStatementWithConstantExprOnlyReturnTheLineOfLast(): vo
...
+                355,
+                370,
+                377,
+                390,
+                392,
+                402,
+                404,
+                414,
+                416,
+                423, // should be 424 as reported by xdebug
```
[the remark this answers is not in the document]
mvorisek: I am working on the fix, the problem is every array item is executable when at least one item is executable - https://3v4l.org/Ij1r9/vld (see VLD tab)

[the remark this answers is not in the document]
mvorisek: @Slamdunk the line `423` is fixed and https://github.com/mvorisek/php-code-coverage/pull/2 merged, please review the whole PR, I will then squash it and undraft

mvorisek:
PR is RTM, I squashed the commits and for easier review I separated "else statement fix" into a separate commit as it affected many test reports

https://github.com/sebastianbergmann/php-code-coverage/compare/858b0f26c8b12ed8f9bf1a610e1470af6950c5da...05f06733e95b70dac27f1d657f1372010978ef78 comparison before squash can help you to understand/follow the static analyser changes made

Slamdunk (commented):
`tests/_files/source_with_heavy_indentation.php` has been changed too much, it's a bit exausting to review now, but I trust @mvorisek work on this.

https://github.com/mvorisek/php-code-coverage/pull/3 instead need to be merged

Slamdunk:
@mvorisek thank you for your patience and explanations.

I approve this PR.

Nevertheless, I now think we are fixing the wrong problem. What `ExecutableLinesFindingVisitor` does is creating an AST-based code coverage analysis, which will always be wrong since xDebug/PCOV doesn't produce and AST-base CC.
I'll open a discussion tomorrow with a feasible solution. In the meantime this can be merged.

sebastianbergmann:
> I approve this PR.

Good to know. However, the type checker result scares me:

```
ERROR: PossiblyUndefinedMethod - src/StaticAnalysis/CodeUnitFindingVisitor.php:317:36 - Method PhpParser\Node\IntersectionType::toString does not exist (see https://psalm.dev/108)
                $types[] = $_type->toString();
```

mvorisek:
> ```
> ERROR: PossiblyUndefinedMethod - src/StaticAnalysis/CodeUnitFindingVisitor.php:317:36 - Method PhpParser\Node\IntersectionType::toString does not exist (see https://psalm.dev/108)
>                 $types[] = $_type->toString();
> ```

@sebastianbergmann I did not change the `CodeUnitFindingVisitor` class, the error is probably due some fix in external CS/CI dep

sebastianbergmann:
Merged manually, thanks.

mvorisek:
@sebastianbergmann would it be possible to tag a new release, please?

Slamdunk:
I'm pouring my energies into https://github.com/sebastianbergmann/php-code-coverage/issues/959 I hope to give a valuable feedback in few weeks

PR #951, opened 2022-10-27
Fix: Remove unused key-value pair

localheinz:
This pull request

- [x] removes an unused key-value pair


localheinz:

Issue #953, opened 2022-11-01
Code coverage changed between 9.2.17 and 9.2.18


kukulich:
**phpunit/php-code-coverage 9.2.17**

![before](https://user-images.githubusercontent.com/260445/199172784-62c1521f-7393-4729-ad11-fd95387719d7.png)

**phpunit/php-code-coverage 9.2.18**

![after](https://user-images.githubusercontent.com/260445/199172791-3a17d9ab-147c-4cb1-a6a8-9d9e6da2c3c8.png)

The problem is that [infection/infection](https://github.com/infection/infection) now think the line is not covered and reports mutants.

With phpunit/php-code-coverage 9.2.17: `10 mutants were not covered by tests`
With phpunit/php-code-coverage 9.2.18: `68 mutants were not covered by tests`

```diff
29) /home/runner/work/BetterReflection/BetterReflection/src/Reflection/StringCast/ReflectionAttributeStringCast.php:27    [M] Concat

--- Original
+++ New
@@ @@
     {
         $arguments = $attributeReflection->getArguments();
         $argumentsFormat = $arguments !== [] ? " {\n  - Arguments [%d] {%s\n  }\n}" : '';
-        return sprintf('Attribute [ %s ]' . $argumentsFormat . "\n", $attributeReflection->getName(), count($arguments), self::argumentsToString($arguments));
+        return sprintf($argumentsFormat . 'Attribute [ %s ]' . "\n", $attributeReflection->getName(), count($arguments), self::argumentsToString($arguments));
     }
```

Slamdunk:
There's a lot to talk about here.

First: IMHO `infection` shouldn't generate mutants for lines that `php-code-coverage` doesn't consider executable.
Maybe this should be "*resolved*" by using `--only-covered` infection's flag, I'm not really sure, but that's none of `php-code-coverage` business

Second: is `'Attribute [ %s ]' . $argumentsFormat . "\n"` and executable line?
I've done some research, and yes it should be because `$argumentsFormat` could undergo a typecasting during Concat that may generate errors (like a `__toString()` class), so indeed we should take this into account

Third: does a fix for the second point resolve the issue completely? No.
Take into consideration this example:

```php
/* 1 */    $var
/* 2 */        =
/* 3 */            'a'
/* 4 */            .
/* 5 */            'b'
/* 6 */    ;
```
[...]

PR #955, opened 2022-11-04
Fix multiline variable in array and with expr not be coverage

kayw-geek:
I think the case need be coverage
Fix #948

### DIFF
--- a/src/StaticAnalysis/ExecutableLinesFindingVisitor.php
+++ b/src/StaticAnalysis/ExecutableLinesFindingVisitor.php
@@ -172,9 +172,36 @@ final class ExecutableLinesFindingVisitor extends NodeVisitorAbstract
                 return [$node->getEndLine()];
             }
 
-            if ($node->items[0] instanceof ArrayItem) {
-                return [$node->items[0]->getStartLine()];
+            $lines = [];
+
+            foreach ($node->items as $index => $item) {
+                if (!$item instanceof ArrayItem) {
+                    continue;
+                }
+
+                if ($index === 0) {
+                    $lines[] = $item->getStartLine();
+
+                    continue;
+                }
+
+                if ($item->key instanceof BinaryOp ||
+                    $item->value instanceof BinaryOp) {
+                    continue;
+                }
+
+                if ($item->key instanceof Node\Expr\Variable) {
+                    $lines[] = $item->key->getStartLine();
+
+                    continue;
+                }
+
+                if ($item->value instanceof Node\Expr\Variable) {
+                    $lines[] = $item->value->getStartLine();
+                }
             }
+
+            return $lines;
         }
 
         if ($node instanceof ClassMethod) {
--- a/tests/tests/Data/RawCodeCoverageDataTest.php
+++ b/tests/tests/Data/RawCodeCoverageDataTest.php
@@ -333,7 +333,7 @@ final class RawCodeCoverageDataTest extends TestCase
         );
     }
 
-    public function testEmtpyConstructorIsMarkedAsExecutable(): void
+    public function testEmptyConstructorIsMarkedAsExecutable(): void
     {
         $file = TEST_FILES_PATH . 'source_with_empty_constructor.php';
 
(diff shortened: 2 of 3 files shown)


sebastianbergmann:
CC @Slamdunk @mvorisek

kayw-geek:
image
Before, variable not be coverage

mvorisek:
please submit this PR into my https://github.com/sebastianbergmann/php-code-coverage/pull/949 PR, the analysis is rewritten a lot

kayw-geek:
> @mvorisek you cannot expect a new contributor to know and inspect work that hasn't been merged yet.
> 
> We either complete and merge your work and ask @kayw-geek to rebase his PR afterwards, or simply integrate the new test case into #949 tests and fix it by ourselves.

It took me half a day to find this problem, and I really want to be a contributor. I will feel honored.

tests/tests/Data/RawCodeCoverageDataTest.php
```diff
@@ -417,6 +417,9 @@ public function testReturnStatementWithConstantExprOnlyReturnTheLineOfLast(): vo
                 226,
                 235,
                 244,
+                250,
+                254,
```
mvorisek, line 421: line 255 is variable as well and reported by xdebug, so it should be marked as executable/covered as well

tests/_files/source_with_multiline_constant_return.php
```diff
@@ -244,4 +244,17 @@ public function Spaceship(): int
...
+        $n = 1;
+
+        return
+            [
+                2 + $n,
+                $n ,
+                2 . $n,
+            ]
+            ;
+    }
```
mvorisek, line 259:
```suggestion
    public function VariableInArray(): array
    {
        $a = 1;
        $b = 2;

        return
            [
                $a
                +
                $b
                +
                $b,
                $a ,
                2
                +
                $a
                +
                $b
                +
                $b,
                $a ,
                $a
                .
                $b
                .
                $b,
                2
                .
                $a
                .
                $b
                .
                $b,
            ]
            ;
    }
```

test with this example and expected output:
```
250,
251,
// line 255 must not be present as not present in xdebug output
257,
259,
260,
263,
265,
267,
268,
// line 269 must not be present as not present in xdebug output
271,
273,
276,
278,
280,
```

@derickr why lines 255 and 269 are not present in xdebug coverage output or is it a bug?

tested on PHP 8.1.12 Win x64 and xdebug 3.1.4 and xdebug 3.2.0RC1

tests/_files/source_with_multiline_constant_return.php
```diff
@@ -244,4 +244,17 @@ public function Spaceship(): int
...
         ;
     }
+
+    public function VariableInArray(): int
+    {
+        $n = 1;
+
+        return
+            [
+                2 + $n,
```
mvorisek, line 254: added extended test with heavy new lines in https://github.com/sebastianbergmann/php-code-coverage/pull/949

`2 + $n` expr itself must not be covered as such expr is unstable and the presence in coverage depends on opcache cache, it makes sense, as `2 + 1` is constant expression that can be evaluated at compile time

kayw-geek: So you mean is the variable in the array should be covered, as well the variable plus a number shouldn't be covered?
This makes me a little confused.

mvorisek: the definition of "executable line" is not simple, but in short, it must address two things:
- when we say about something it is "an executable line", it must be always in xdebug coverage
- it must be stable, see https://github.com/sebastianbergmann/php-code-coverage/issues/889

kayw-geek: Thanks, I got it

kayw-geek:
@mvorisek Excuse me, Could you review this PR again. Thanks 🙏

src/StaticAnalysis/ExecutableLinesFindingVisitor.php
```diff
@@ -172,9 +172,40 @@ private function getLines(Node $node): array
...
+                    continue;
+                }
+
+                if ($item->key instanceof BinaryOp ||
+                    $item->value instanceof BinaryOp) {
+                    continue;
+                }
+
+                if ($item->key instanceof Node\Expr\Variable ||
+                    $item->key instanceof Node\Expr\ClassConstFetch
```
mvorisek, line 194: I would say this will return the unstable behaviour described in https://github.com/sebastianbergmann/php-code-coverage/issues/889

kayw-geek:
> 

So, your meaning was I need to use the method in the #939 to fix these unstable problems?

kayw-geek:
> see
> 
> https://github.com/sebastianbergmann/php-code-coverage/blob/9ffa0eb5e8440d4f008654c5436aa8dd24f47306/tests/tests/Data/RawCodeCoverageDataTest.php#L359
> 
> test and explanation in [#953 (comment)](https://github.com/sebastianbergmann/php-code-coverage/issues/953#issuecomment-1310831008)
> to fix the #953 you need to statically analyse the local variable scope (范围) for each evaluated (评价) node to be able to determine (确定) if variable was optimized out at compile (编译) time (const expr) or not

So, your meaning was I need to use the method in the https://github.com/sebastianbergmann/php-code-coverage/pull/939 to fix these unstable problems?

mvorisek:
you need to evaluate every (sub) AST tree for const expr and track variable scope for each, like done by php optimizer or other static analysers like phpstan

in a meantime, please also review https://github.com/sebastianbergmann/php-code-coverage/pull/949, it will help you to understand some edge cases

PR #956, opened 2022-11-08
Improved the Cobertura coverage report

jhoffland:
The format of the Cobertura report did not match the [Cobertura document definition](http://cobertura.sourceforge.net/xml/coverage-04.dtd). This also resulted in tools like Jenkins that were not able to interpret the report.


sebastianbergmann:
If this is indeed a bugfix then it should go into the `9.2` branch.

Issue #960, opened 2022-11-19
New body font-size is way too big


LordSimal:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 9.2.19
| PHP version               | 8.1.10
| Driver                    | Xdebug
| Xdebug version  | 3.1.5
| Installation Method       | Composer
| Usage Method              | PHPUnit
| PHPUnit version | 9.5.26

I just updated to the current latest version, regenerated the Code coverage page and was very confused why the text was so big.

<img width="1728" alt="image" src="https://user-images.githubusercontent.com/9105243/202868188-a45a23d7-9d95-4fb8-9cda-3cafbe1a5b90.png">

Just compare the URL font-size compared to the rest of the text.

This especially is way to wasteful when looking at code lines

<img width="1728" alt="image" src="https://user-images.githubusercontent.com/9105243/202868231-ae3683ed-9b24-4fdb-b86a-4573fb1453c2.png">

As you can see here without any zooming I can only see 18 lines of code.

Compared to the old (`front-size: 1em`) I am able to read 37 lines of code in 1 page.

What was the reason to change the default body font-size to 2em in [HERE](https://github.com/sebastianbergmann/php-code-coverage/commit/aa0971fe8c9de5423670c6cd5954f27b89274f2f)?

gemal:
yes it's almost unusable now :(

sebastianbergmann:
I am not a frontend developer, sorry. Setting `font-size` to `2em` in aa0971fe8c9de5423670c6cd5954f27b89274f2f was motivated by the fact that I wanted the code coverage report to be readable (for me) without having to zoom.

LordSimal:
Thanks for the fix and the quick release 😄 

Regarding your problem:
Depending on your monitor size, resolution and the UI scaling setting in your OS the text can definitely be very small. I can understand that.
The proper solution to this problem would be to add a body font-size changer element (maybe at the top right) with `+` and `-` buttons. If you want I can provide a PR for that.

sebastianbergmann:
> If you want I can provide a PR for that.

Yes, please.

PR #961, opened 2022-11-20
add body font-size changer

LordSimal:
Refs: https://github.com/sebastianbergmann/php-code-coverage/issues/960

This adds 4 Buttons at the top right of the coverage page to let the user choose which kind of font-size they prefer.

Currently there is
- small (`0.75em`)
- medium (`1em`) which is default Browser behavior (16px)
- large (`1.5em`)
- xlarge (`2em`)

The way how this is saved across requests is done via localstorage and therefore currently requies JS to be executed.

This also introduces a small CLS jump. This could be mitigated via implementing it via cookies and adding the correct data-attribute to the body in PHP beforehand.


sebastianbergmann:
Thanks! Can you send this against the `9.2` branch, please? A tweak such as this is safe to to make in a stable branch.

Issue #959, opened 2022-11-16
Line Coverage: mark each line of the Branch as executable


Slamdunk:
In the past year I had deluded myself that an ever curated `ExecutableLinesFindingVisitor` could eventually reach an acceptable quality over what to consider effectively an exec line or not.

That will never be the case: xDebug and PCOV are not AST-based, so neither `ExecutableLinesFindingVisitor` can be so.

This is even more problematic considering that PHP/xDebug/PCOV just lie.

Take this example, where *no* bugs of `php-code-coverage` is involved:
![https://github.com/infection/infection/issues/1750](https://user-images.githubusercontent.com/152236/200500491-df8b496b-356d-4757-9a24-65da61d6beb0.png)

Lines `12`, `13`, `14`, `15` and `16` have been executed for sure by the PHP engine, but the CC driver doesn't consider them as executed.

A solution can be to go the other way around of what we've done so far: consider each line of a branch as executable.

*Disclaimer*: I'm not talking about Branch/Patch coverage provided by xDebug with `XDEBUG_CC_BRANCH_CHECK`. That's a good feature, but pure line coverage is still the fastest and most supported CC methodology available everywhere, so that's where I want our attention to focus on.

The idea is: treat every line as executable, split them each into their *branch* of execution, and consider all lines of that branch as green if any of them is green according to xDebug/PCov.

Let's see this example:

```php
$var
    =
    1
;

if ($var === 2) {
    $var
        .=
        2
    ;
}

var_dump(
    $var
);
```

A run with `xdebug_start_code_coverage(XDEBUG_CC_UNUSED|XDEBUG_CC_DEAD_CODE)` gives this result:

```php
$var                 //  1
    =
    1
;

if ($var === 2) {    //  1
    $var
        .=
        2            // -1
    ;
}

var_dump(            //  1
    $var             //  1
);
```

The AST-based coverage is complex, but there are only 2 branches here, really easy to spot: outside and inside the `if`.
Knowing this, we can run this short checklist:
1. Is there **any** line reported as `1` outside the `if`? Yes; then all the lines of that branch are executed
2. Is there **any** line reported as `1` inside the `if`? No; then no lines of that branch are executed

The final report can be reported as:

```php
$var                // green
    =               // green
    1               // green
;                   // green

if ($var === 2) {   // green
    $var            // RED
        .=          // RED
        2           // RED
    ;               // RED
}

var_dump(           // green
    $var            // green
);                  // green
```

[...]

Issue #945, opened 2022-10-11
Coverage not working on closures in PHPDBG


Ocramius:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 9.2.17
| PHP version               | 8.1.10
| Driver                    | PHPDBG
| PCOV version (if used)    | 1.0.11
| Xdebug version (if used)  | 
| Installation Method       | Composer
| Usage Method              | PHPUnit
| PHPUnit version (if used) | 9.5.25

As investigated in https://github.com/infection/infection/issues/1736, coverage for closures doesn't seem to be picked up correctly, when using PHPDBG.

Specifically, `PCOV` works as expected:

![Selection_1052](https://user-images.githubusercontent.com/154256/195045584-0162f874-bdf3-4f10-a770-a8683debdba2.png)

A coverage report done with `phpdbg -qrr` will instead produce partial coverage:

![Selection_1051](https://user-images.githubusercontent.com/154256/195045738-6a8925ee-f87e-44ed-8dbb-4c393b0926f0.png)

## Reproducible example

```php
<?php

declare(strict_types=1);

final class Foo
{
    /** @return callable(): int */
    function make(): callable
    {
        return static function (): int {
            return 1;
        };
    }
}
```

```php
<?php

declare(strict_types=1);

require_once __DIR__ . '/../../src/Foo.php';

/** @covers Foo */
final class FooTest extends \PHPUnit\Framework\TestCase
{
    function testClosure(): void
    {
        self::assertSame(1, (new Foo())->make()());
    }
}
```

### PHPDBG output

```
❯ phpdbg -qrr ./vendor/bin/phpunit --coverage-html=coverage --filter=Foo
PHPUnit 9.5.25 #StandWithUkraine

.                                                                   1 / 1 (100%)

Time: 00:00.140, Memory: 50.00 MB

OK (1 test, 1 assertion)

Generating code coverage report in HTML format ... done [00:00.812]
```

![Selection_1053](https://user-images.githubusercontent.com/154256/195046392-ce8d075c-a976-49d5-9685-b567795bfa19.png)

### PCOV output

```
❯ ./vendor/bin/phpunit --coverage-html=coverage --filter=Foo
PHPUnit 9.5.25 #StandWithUkraine

.                                                                   1 / 1 (100%)

Time: 00:00.053, Memory: 40.00 MB

OK (1 test, 1 assertion)

Generating code coverage report in HTML format ... done [00:00.708]
```

[...]

PR #964, opened 2022-11-24
Group line code-coverage by branch

Slamdunk:
Fix https://github.com/sebastianbergmann/php-code-coverage/issues/959

- [x] Move tests over `ExecutableLinesFindingVisitor` from `RawCodeCoverageDataTest` to its own unit test
- [x] Inline test data expectations to ease readability and maintainability
- [x] Make test data expectations relative, so new test data can be added before or after the current ones without the need to rewrite other test data
- [x] Save an associative array of `line <=> branch index`
- [x] Add a test for every [Control Structure](https://www.php.net/manual/en/language.control-structures.php) and language definitions (Functions and Classes)
- [x] Intersect `line <=> branch index` association with CC driver's output to obtain the final CC
- [x] Test multiple CC data merge
- [x] Test PR over existing code bases

#### Before

image

#### After 

image

As you can see, the difference can be quite high on line coverage: the code base in the shown here uses for example a lot of array declarations; prior to this PR, a single 1.000 array would have produced lines of CC only on non-scalar nodes; this PR instead marks all the 1.000 as executed (or not).

But as we know, line CC is convenient, quick and reliable only as a quantitative result.
Functions/Methods/Classes/Traits coverage has a much higher qualitative value, and this PR has very little effect on results. The difference in these cases are **correct** and expected: current `9.2.19` release has bugs that this PR fixes.

### Special cases

Some PHP structure cannot be deterministically handled well. For example:

```php
return match($var) {
    MyEnum::A => 1,
    MyEnum::B => 1,
};
```

This code will always depend on autoload runtimes, but in contrast to https://github.com/sebastianbergmann/php-code-coverage/pull/892 this will also always produce 4 executable lines from CC drivers.
This PR favour determinism and stability over CC precision, so for the example provided a *single* branch is reported, and thus all the 4 lines are going to be marked as covered if `return` has any hit.

If the user needs precise and reliable feedback over CC, only Mutation Testing can help; in the `match` example it would be `MatchArmRemoval` mutators from Infection, see https://infection.github.io/guide/mutators.html#Removal-Mutators


mvorisek:
IMHO we should not build ranges like `($node->getStartLine(), $node->getEndLine())` but instead build tree of all possible coverage points (see 2. below)

1. detect coverage points like now - these are always fine
2. detect all possible coverage points no matter if const expr or not (ie. to match xdebug coverage /wo any opcache/optimization enabled)
3. then compute dependent/branched coverage points
   - if one point from group is covered by xdebug, then mark all points as covered
   - if one point from group is not covered (but present) by xdebug, then mark all points as uncovered
   - you need to short circuiting correctly like `expr && expr` or `a ? b : c`
4. the output must never cover more lines than in 2.
5. must fix https://github.com/sebastianbergmann/php-code-coverage/issues/889

I am not sure if this can solve every situation, imagine code like:
```
$a =
mt_rand() === 0
?
'a' + 'a'
:
'b' + 'b'
;
```

the `'a' + 'a'` must NOT be marked as covered until really executed (and the chance of it is minimal, as 63 bit random number is very unlikely exactly 0)

if `mt_rand() === 0` would be `Cl::const` instead, there are simply no data to tell which ternary path was taken

also please note

```
try {fxA()
.
'b'
;} catch (\Throwable) {}
```

must not mark `'b'` as covered when `fxA()` will throw even if upper branches (try block, function block) are covered

ping me once the PR is done, I can then comment on the impl. and find another test cases

Slamdunk:
> IMHO we should not build ranges like `($node->getStartLine(), $node->getEndLine())` but instead build tree of all possible coverage points (see 2. below)

I think the opposite: line code-coverage is **by definition** wrong, but that's all we have. Starting from the AST is just a waste of time: let's go strait to lines without worrying about the tree.

> 1. detect coverage points like now

What to you mean by "now", `9.2` branch or this PR?

> the `'a' + 'a'` must NOT be marked as covered until really executed

That's obvious, why did you doubt it?

> if `mt_rand() === 0` would be `Cl::const` instead, there are simply no data to tell which ternary path was taken

I don't get what you are trying to say: if you format the ternary correctly, we know which path has been taken, and we can mark as green only the taken path.

> ```php
> try {fxA()
> .
> 'b'
> ;} catch (\Throwable) {}
> ```

Let's be clear: line code-coverage is flawed by design. I couldn't care less about this example: if you are collecting the code-coverage with the line CC driver, it is required that you format your code correctly:

```php
try {
    $var = fxA() . 'b';
    $var .= 'c';
;} catch (\Throwable) {}
```
If `fxA()` throws an `Exception`, `$var = fxA() . 'b';` is going do be marked as executed, while `$var .= 'c';` not.
If a developer likes to write everything in one line, it's not this library's fault if the result is meaningless.

There is no solution that can fix every edge case. Let's focus on providing the best value for the sensible developers.

mvorisek:
> I think the opposite: line code-coverage is by definition wrong, but that's all we have. Starting from the AST is just a waste of time: let's go strait to lines without worrying about the tree.

> I don't get what you are trying to say: if you format the ternary correctly,

> If a developer likes to write everything in one line, it's not this library's fault if the result is meaningless.

all about the same topic - I agree it is a convention to code that is usually "friendly" to your approach, but I very strongly disagree we should do it that way.

> the 'a' + 'a' must NOT be marked as covered until really executed (and the chance of it is minimal, as 63 bit random number is very unlikely exactly 0)

> if mt_rand() === 0 would be Cl::const instead, there are simply no data to tell which ternary path was taken

⚠️ this is quite common non-rare example and you must handle it, showing some code path/branch which was never taken as covered is very bad and very hard to debug from user side

> > 1. detect coverage points like now
>
> What to you mean by "now", 9.2 branch or this PR?

Currently, for multiline statement/expression, we save only 1 coverage line/point when it is really emit by php/xdebug. This should stay IMHO. If something is a part of an expression/statement branch should be based on AST not an expr. start/end line, this has no physical meaning, no such data is emit by xdebug.

Example:
```
$v = <<<EOF
text
EOF
.
'b';
```

you should build about this tree for it:
```
assign (l1)
-> binary concat
-> -> string scalar (l2)
-> -> string scalar (l5)
```
then you know variable assign op can be covered if and only if the inner expression aka branch has completed, so, the l2 and l5 can be marked as covered if l1 was covered

but when the the code will be formatted like:
```
$v2 = 'x'; $v = <<<EOF
text
EOF
.
'b';
```
the tree should be:
```
assign (l1)
-> string scalar (l1)
assign (l1)
-> binary concat
-> -> string scalar (l2)
-> -> string scalar (l5)
```
and you cannot transfer any coverage information from assign l1 into l2/l5 as l1 source cannot be used as the coverage can be comming from unrelated statement

Slamdunk:
> :warning: this is quite common non-rare example and you must handle it, showing some code path/branch which was never taken as covered is very bad and very hard to debug from user side

I sincerely have no idea what you're talking about here. Ternary operator gets 4 line coverage from CC drivers:

```php
$var            // 1
    =
    true        // 1
    ?
        'b'     // 1
        :
        'c'     // -1
    ;
```
Assignment and if-condition are on the same logical branch, and then we have the 2 branches `b` and `c`.
We redact 3 distinct branches, and look for their respective usages.
Constant expr and opcache optimisation don't matter here: CC driver will always report the result for at least the first line of each branch.

> and you cannot transfer any coverage information from assign l1 into l2/l5 as l1 source cannot be used as the coverage can be comming from unrelated statement

Can you provide a real example? As far as I can tell, L1 and L2/L5 are coupled: you cannot execute one without the other.

I am already taking into account `return` \ `exit` \ `throw` \ `goto` statements as well as `function` and `method` calls to divert a new branch.

tests/tests/Data/RawCodeCoverageDataTest.php
```diff
@@ -264,295 +262,6 @@ public function testRemoveCoverageDataForLines(): void
...
-                32,
-                33,
-                35,
-                40,
-            ],
-            array_keys(RawCodeCoverageData::fromUncoveredFile($file, new ParsingFileAnalyser(true, true))->lineCoverage()[$file])
-        );
-    }
-
-    public function testHeavyIndentationIsHandledCorrectly(): void
```
mvorisek, line 330: This test is very hardly crafted test, please keep it. Unit tests != functional tests, this test is more like a functional test and tests many specific combinations. `source_with_multiline_constant_return.php` test is also very important, at least keep these two, but I would prefer no single test deleted (if not migrated 1:1).

Slamdunk: I am aware of your previous work on `RawCodeCoverageDataTest`; I know the effort you put on it, also because I carefully crafted and reviewed it *two* times before you started contributing here: very time consuming and boring.

But all tests currently in `9.2` in `RawCodeCoverageDataTest` related to `ExecutableLinesFindingVisitor` must go:

1. The underlying logic has been completely changed: the data sets and the related comments just don't make sense anymore
2. The update needed to get the test pass marks all lines as executable, and wouldn't test how they are grouped by branch, so it wouldn't have any value as a regression test
3. You know how much of a PITA they currently are in `9.2` to both write and review, they are unbearable to maintain

Don't look at what we are losing, look at what we gained: this PR couldn't have been written without the precious knowledge and experience we got from working on it.

All the value we poured in our previous PR is still here: the (little) value you don't see anymore in raw data sets has been consumed in the form of a new architecture.

mvorisek: I support the in place test annotation, but I the original data must be kept/tested 1:1. They tests many issues, syntetic or found during development and prevent https://github.com/sebastianbergmann/php-code-coverage/issues/942 which strictly requires executable lines to be no larger than maximum subset of real executable lines emit by coverage driver.

> All the value we poured in our previous PR is still here: the (little) value you don't see anymore in raw data sets has been consumed in the form of a new architecture.

That is what I am asking for, but why the original tests does not pass if I revert them?

Slamdunk: > why the original tests does not pass?

Because the logic changed.

> prevent #942 which strictly requires executable lines to be no larger than maximum subset of real executable lines emit by coverage driver.

The logic changed and this requirement is no longer valid.

This PR extends the green/red lines over all the AST node a driver passes.

Slamdunk:
@sebastianbergmann @dvdoug this is ready for review now

tests/tests/Data/RawCodeCoverageDataTest.php
mvorisek: but this will massively return https://github.com/sebastianbergmann/php-code-coverage/issues/942 bug

I still strongly belive you should not emit more coverage points than is normally emit by xdebug.

mvorisek: I will definitely test this PR. but please remind me why multiple green lines for for ex. multiline string are better than one?

I do not see any reason we should report/count more than one line for I source executable point/data. This is how Istanbul JS (and other coverage software) work.

In the description you mentioned (in real project) the coverage changed 93.65% to 96.11%. But what number is better? The first one has industry standard (+- xdebug/pcov/IstanbulJS) percentage and better informational value. If a string with 1M lines is added to a project, it must to skew the coverage by more than 1 executable line.

I understand the https://github.com/sebastianbergmann/php-code-coverage/issues/959 problem, I belive reasoned implied coverage points by executed branches can be a solution, but you should not report more lines than meaningfull (+- lines without opcache/non-costs expressions). So old tests should/must still pass and if you want to extend the coverage data for more lines, make it optional.

Also, as coded now this cannot be considered as a bugfix.

dvdoug: Haven't looked at the PR in depth yet, but I think I would tend to agree that a hardcoded multiline string, or hardcoded multiline array should only count as 1 executable line - the assignment part is definitely executable code, but the contents thereafter are really just an opaque data blob once parsed - the choice of using '\n' or an actual newline in a string is irrelevant to which parts of the code are executed or not and it would be surprising to have code coverage numbers change if someone e.g. added an extra sentence into an email body.

Where a string has interpolation that calls into methods or an array has calculated values, then those particular lines I would expect to be identified as executable individually

Slamdunk: First things first, I managed to handle strings better, and now multiline strings have only 2 lines counted, the first and last ones; in between these two, no lines are marked anymore as executable.

But I urge you to focus on what really important this PR is finally achieving: reliability.

Of course I agree that only meaningful lines should produce CC, but what do we do when this cannot be done reliably? There's a scale of priorities we need keep in mind:

1. Non executed executable lines must be marked as red
2. Executed executable lines must be marked as green
3. CC reports must be deterministic and not influenced by the tools that produce or use them
4. Only meaningful lines should be reported

I'll do my best to achieve every listed point, but we *need* to sacrifice [4] *when* [1], [2] and [3] cannot be simultaneously satisfied.

The proposed solution achieve first three points *all* of the times and all four points *most* of the times, with a relative simpler code.

mvorisek: You did not answer my question, what is the physical meaning (of emiting coverage for first and last string line, for empty lines, operator symbols, ...)?

In the PR description you reason this by:

> But as we know, line CC is convenient, quick and reliable only as a quantitative result.
Functions/Methods/Classes/Traits coverage has a much higher qualitative value, and this PR has very little effect on results. The difference in these cases are correct and expected: current 9.2.19 release has bugs that this PR fixes.

So basically you say line coverage has little imformation value (qualitative value). This is not true, as line coverage is normally the primary value reported by tools like CodeCov.

Also this PR has exactly oposite then little effect on results as 93.65% vs. 96.11% line coverage is huge change.

In the latest release we track executable lines well and #953/#959 is basically not even a bug, it is a feature request.

I belive you need to change the result of your branch analysis. We both know quite well when/where xdebug emit coverage points and we know when it does not (always or sometimes due opcache). The result I expect from branch analysis PR is an update of the original executable analysis to analyse two kinds of executable lines types - certain and maybe lines - certain lines should match the original result and maybe lines should match the original result assuming the analysed expression was not a const expr. Then (like you do now), reason the maybe lines using the certain ones. This will fix #953 and keep everyone happy.

src/StaticAnalysis/ExecutableLinesFindingVisitor.php
```diff
@@ -57,285 +30,325 @@
 final class ExecutableLinesFindingVisitor extends NodeVisitorAbstract
```
mvorisek, line 30: I am opening another thread here to discuss the reasoning (result validity) only.

code:

```
    public function testBranchCoverage(): void
    {
        $a =
            mt_rand() === 0
            ?
            'a'
            .
            'a'
            :
            'b'
            .
            'b'
            ;

        try {fxA()
        .
        fxB()
        ;} catch (\Throwable $e) {}

        $b = true ? 'a' : fxA()
            .
            'c';

        $c = [


            'x'



        ];


        [


            'x'



        ];

        if (mt_rand() === 0) {
            var_dump($a . $b . count($c));
        }
    }
```

image

was whole marked as red, is it really possible it was executed? (is public and starting with test, other testXxx methods were executed/covered)

tests/tests/Data/RawCodeCoverageDataTest.php
Slamdunk: > You did not answer my question, what is the physical meaning (of emiting coverage for first and last string line, for empty lines, operator symbols, ...)?

My previous comments answer the question: the physical meaning (point [4] of my list) has no value when CC is unreliable

> So basically you say line coverage has little imformation value (qualitative value). This is not true, as line coverage is normally the primary value reported by tools like CodeCov.

Spread and adoption !== value

Line Code Coverage has little value because it can be easily cheated, both on purpose and unintentionally.
Line Code Coverage is good only for finding uncovered lines (red ones), but a single executed line can have many code branches and lead to many code paths, so seeing green doesn't mean your tests are good.

Only Mutation Testing can prove the tests are good. Line Code Coverage is still important though to ease the work and focus of MT on diffs.

> Also this PR has exactly opposite then little effect on results as 93.65% vs. 96.11% line coverage is huge change.

The amount of uncovered lines staid the same, so no change at all in my PoV.

> We both know quite well when/where xdebug emit coverage points and we know when it does not (always or sometimes due opcache). The result I expect from branch analysis PR is an update of the original executable analysis

Autoload runtimes are unpredictable, as well as opcache ones, so no one can predict where a CC driver is going emit a CC point for a particular test run.
This is way this PR cannot be an update of the current `9.2` implementation

src/StaticAnalysis/ExecutableLinesFindingVisitor.php
Slamdunk: Please elaborate further, I don't understand the environment nor you question

tests/tests/Data/RawCodeCoverageDataTest.php
mvorisek: > My previous comments answer the question: the physical meaning (point [4] of my list) has no value when CC is unreliable

really? Please provide an example of unreliable, more precisely "unstable" line, ie. line that is sometimes present in xdebug as 1, sometimes as 0.

> Autoload runtimes are unpredictable, as well as opcache ones, so no one can predict where a CC driver is going emit a CC point for a particular test run.

AFAK this is not so much about CC drivers, but more about php-src, as php-src tries to compile const expr. code - but the executable lines are emit on quite well known places.

Slamdunk: > really? Please provide an example of unreliable, more precisely "unstable" line, ie. line that is sometimes present in xdebug as 1, sometimes as 0.

https://github.com/sebastianbergmann/php-code-coverage/pull/892

> AFAK this is not so much about CC drivers, but more about php-src, as php-src tries to compile const expr. code - but the executable lines are emit on quite well known places.

Exactly, and that's change nothing for us

src/StaticAnalysis/ExecutableLinesFindingVisitor.php
mvorisek: I have added your PR branch as a dep, added the test code I posted to one phpunit test class...

It was completely red as there was not assertion (expected behaviour).

image

here are the objections I already posted

you must collect certain/maybe coverage points together with AST, and reason the maybe points based on AST node types/behaviour

tests/tests/Data/RawCodeCoverageDataTest.php
mvorisek: You have my support to show coverage support for unstable lines, we know exactly where the coverage point is emit (/wo opcache). This PR should do the analysis and if it can always reason the coverage point, it should add an executable line.

src/StaticAnalysis/ExecutableLinesFindingVisitor.php
Slamdunk: I don't know what's going on as it seems to me your are not using this PR correctly:

1. This PR doesn't mark your `L134` as executable
2. This PR does mark `L171`, `L174` and `L178` as executable and executed

I don't care about `L159`: if you put the `else` part of the ternary operator in a new line it become red again; users must use a proper coding style fixer to expect Line CC to be meaningful.

But, assuming `fxA()` is throwing an exception (which I don't know since you didn't provide the full code, but only a screenshot) I can confirm that `L154` must be green even though CC drivers are reporting it as not executed.
The same will go for method concatenation on objects:
```php
$myObject
    ->this_method_throws_an_exception()
    ->this_method_will_be_reported_as_executed_anyway()
;
```

The reason is that what you propose is impossible to be reliable, accurate and maintained.
For example:

```php
function foo($var) {
    return
        $var
        .
        'b'
        ;
}
function bar($var) {
    return
        $var
        .
        MyClass::FOO
        ;
}
```

Should we consider `$var` line as executable or not? The answer is it depends if `$var` is an object or not, and you can't know it only by parsing the code.
Let's compare current behaviour with this PR:

| Scenario | `9.2` report | PR report |
| --- | --- | --- |
| `foo` not called | :red_circle: `'b'` line red | :red_circle: all lines red |
| `foo` called smoothly | :heavy_check_mark: `'b'` line green | :heavy_check_mark: all lines green |
| `foo` called, `$var` object typecast raises an Exception | :red_circle: `'b'` line red | :heavy_check_mark: all lines green |
| `bar` not called | :red_circle: `MyClass::FOO` line red | :red_circle: all lines red |
| `bar` called smoothly | :heavy_check_mark: `MyClass::FOO` line green | :heavy_check_mark: all lines green |
| `bar` called, `$var` object typecast raises an Exception | :heavy_check_mark:/:red_circle: `MyClass::FOO` reported either red or green, depending on autoload runtimes | :heavy_check_mark: all lines green |

This PR is consistent, and where it fails to be precise on `'b'` and  `MyClass::FOO` lines you can enable Mutation Testing to check your tests are good enough.

mvorisek: > I don't know what's going on as it seems to me your are not using this PR correctly:

repro code is here: https://github.com/atk4/ui/commit/5b9438914589de9feb22a22e1f74382a07572909

`L159` no, bad formatting must never even imply wrong (green or even red) coverage, If you will listen to me, and collect certain/maybe executable points together with AST node types, you can apply behavioural reasoning and achive valid output, this is also how phpstan and other tools handle AST

`fxA()` throws undefined function - https://3v4l.org/LlE6W

```
$myObject
    ->this_method_throws_an_exception()
    ->this_method_will_be_reported_as_executed_anyway()
    ->this_method_will_be_reported_as_executed_anyway()
;
```

again, this is a big problem, coverage must not lie

> Should we consider $var line as executable or not? The answer is it depends if $var is an object or not, and you can't know it only by parsing the code.

again, you cannot simply work with topmost AST node start-end range, you must collect executable points

example:

```
    return
        $var
        .
        'b'
        ;
```

`$var` and `'b'` should be analysed as maybe coverage points in an unconditional expr AST node and if the following statement (or an end of method/file or owning expr/branch) is reached (by xdebug data), only then you can reason these maybe coverage points (as covered/uncovered)

> table

(untested) I belive the problem with var is only when `$var` is const expr, a func param is never known upfront, thus `$var` coverage is reported by xdebug (but please, focus on the aka `L159` problem, not these unimportant details)

> This PR is consistent, and where it fails to be precise on 'b' and MyClass::FOO lines you can enable Mutation Testing to check your tests are good enough.

but it introduces many other problems and invalid output data

Slamdunk: > (but please, focus on the aka L159 problem, not these unimportant details)

> but it introduces many other problems and invalid output data

At this point we can only conclude we have different priorities: you consider more important a sometimes 100% correct report, while I consider more important an always 99% correct one.

I don't think this conversation can reach any better conclusion; the maintainers will choose what they prefer.

Have a nice day.

Slamdunk: > again, you cannot simply work with topmost AST node start-end range, you must collect executable points

BTW, this is exactly what we can't do.
For non executed files, this library has no exec points at all, but it must produce a report identical to what it would be with an execution

mvorisek: I do not want to be manipulated at the first place and want to keep focus only on relevant things.

Yes, I expect 100% correct report. We shared this goal until you created this PR and the way I validated the output is by generating reports from more than 10k+ files and comparing the data againts real xdebug output at different PHP versions and PHP configurations. This output is asserted by the latest/`9.2.19` tests.

Saying "sometimes 100% correct report" is wrong and misleading. The report is always 100% correct. In quite small edge cases - compiled constant expressions (/w opcache enabled) at compile time are not present as not present in source/xdebug data. This can be addressed by proper AST branch analysis, but such improvement must never ever imply wrong coverage data.

Have a nice day too and I would be happy if we can agree on this and instead of replacing the correct actual/`9.2.19` executable analysis introduce another analyser, which will take these valid/stable/always correct data and compute optional branch analysis (for mutation testing) based on them. Everyone will be happy :)

Slamdunk: > In quite small edge cases - compiled constant expressions (/w opcache enabled) at compile time are not present as not present in source/xdebug data. This can be addressed by proper AST branch analysis

You can't solve #892 and foo/bar case from https://github.com/sebastianbergmann/php-code-coverage/pull/964#discussion_r1038268222 with AST analysis only.

If you can, code it to prove it please

> instead of replacing the correct actual/9.2.19 executable analysis

Except that it isn't correct because structurally flawed as proven in https://github.com/sebastianbergmann/php-code-coverage/pull/964#discussion_r1038268222 and this it is only sometimes 100% correct

mvorisek: > You can't solve https://github.com/sebastianbergmann/php-code-coverage/pull/892 and foo/bar case from https://github.com/sebastianbergmann/php-code-coverage/pull/964#discussion_r1038268222 with AST analysis only.

for

```
<?php

function u() {
    return (string) mt_rand();
}
function x() {
    return // line 7
    u().
    1;
} // line 10
```

xdebug produces (uncached):

```
Array
(
    [4] => 1
    [5] => -2
    [8] => 1
    [9] => 1
    [10] => -2
)
```

and (cached):
```
Array
(
    [4] => 1
    [8] => 1
    [9] => 1
)
```

for `u` function declaration commented out (`u` function undefined) xdebug produces:

```
Array
(
    [8] => 1
    [9] => -1
)
```

for

```
<?php

function x() {
    return
    '1'.
    1; // line 6
}
```

it produces:

```
Array
(
    [6] => 1
)
```

so I would say it can be 100% correctly reasoned, as if an expression in a statement is const expr, the last expr line is (thankfully) always emit :)

the same for:
```
<?php

function foo($var) {
    return
        $var
        .
        'b' // line 7
        ;
}
```

xdebug produces:

```
Array
(
    [7] => 1
)
```

for (`$var` swapped/last):

```
<?php

function foo($var) {
    return
        'b'
        .
        $var
        ;
}
```

xdebug outputs (thankfully) the same/last expr line:

```
Array
(
    [7] => 1
)
```

Slamdunk: > the same for:
> 
> ```php
> function foo($var) {
>     return
>         $var
>         .
>         'b' // line 7
>         ;
> }
> ```
> 
> xdebug produces:
> 
> ```
> Array
> (
>     [7] => 1
> )
> ```

Exactly, and that's WRONG!

if `$var` is an object, line 5 becomes an executable line too, but you can't know it with a raw CC collection over an unexecuted function, so processing uncovered files would produce a different report from executed files, and that's unreliable

mvorisek: 🙏 understand `$var` variable fetch expression is not an executable line by php-src, thus it is not reported, this is also how other software report coverage lines...

I did an experiment to see if line 5 can become an executable line.

for:

```
<?php

function foo($var) {
    return
        $var
        .
        'b' // line 7
        ;
}

foo(new class () { public function __toString() { return (string) mt_rand(); } });
```

xdebug produces:

```
Array
(
    [7] => 1
    [11] => 1
)
```

with `opcache.enable_cli=0` I get:

```
Array
(
    [7] => 1
    [9] => -2
    [11] => 1
)
```

I am not strictly againts adding a variable fetch expression as an executable line if reasoned correctly, but I am not sure if mutation testing requires it as native xdebug never reported such line (at least not on my side, not "unstable"). I belive the reason is php-src/xdebug reports executable line for binary op, not for the elementary expressions.

In my previous post (and this too), I prooved the topmost expression in a statement does produce at least 1 coverage line. This is enought to reason all executable lines even the ones that do not produce any coverage.

Ocramius:
FWIW, while waiting for this to land, I've raised https://github.com/infection/infection/issues/1773 to propose that `9.2.18` and newer should become incompatible with `infection/infection` until @Slamdunk's work lands.

Slamdunk:
PR is ready to review, we are waiting for @sebastianbergmann feedback.

Con of this PR: some edge case lines that previously were reported either green or red depending of various circumstances, now are always marked as green.

Pros of this PR: reports are now deterministic and consistent, and gives Infection the feedback it needs to always work properly

sebastianbergmann:
I won't have time to look into this (deeply enough) for at least a couple of weeks. But I trust @Slamdunk and @Ocramius :-) If the two of you tell me that this is good to go then I will merge it and roll a release.

Slamdunk:
@Ocramius I guess you don't time to review this either, but I'm more than happy to open demonstrative PRs on any project you're working on with this patch active, so you can check whether it fulfils your expectations or not.

Just throw here the links

Ocramius:
@Slamdunk I'll be able to check it later today, but I'm unfamiliar with the architecture/codebase, so I'm unsure if my review will be useful.

romm:
@Slamdunk I'd be happy to test the patch as well; what would be the procedure? Do I just need to apply this PR's patch locally, and run Infection (where I first saw issues with coverage)?

Slamdunk:
> Do I just need to apply this PR's patch locally, and run Infection

@romm exactly: just discard `tests/` diff as it doesn't land on releases

sebastianbergmann:
> @Slamdunk I'll be able to check it later today, but I'm unfamiliar with the architecture/codebase, so I'm unsure if my review will be useful.

You don't have to review the code changes themselves, "only" the effect they have on the code coverage report in your projects.

romm:
Hi @Slamdunk I just tested it out on [Valinor](https://github.com/CuyZ/Valinor), everything ran smoothly and the MSI went from 89% to (!!) 97%.

Is there anything else I can do to help?

Slamdunk:
> Is there anything else I can do to help?

Yes, check that the remaining 3% is actually untested by your test suite.

romm:
> Yes, check that the remaining 3% is actually untested by your test suite.

It is, that's mostly escaped mutants, the few remaining are indeed not covered by tests.

mvorisek:
@Slamdunk looking into the Valinor data - line like https://dashboard.stryker-mutator.io/reports/github.com/CuyZ/Valinor/master#mutant/Definition/Exception/ClassTypeAliasesDuplication.php is an integer, ie. not an executable, it is a scalar. Such scalar is not reported by any industry standard coverage software like IstanbulJS. In previous discussion with you I have also landed several reasons why more artifical coverage lines might be not wanted.

So I am asking you again. Can you make this optional and based on the originally well defined/tested executable line data?

User can then opt in with a configuration option directly or use https://github.com/sebastianbergmann/phpcov to convert the data using CLI.

Ocramius:
@mvorisek the line you highlighted is an integer **expression**, which is indeed a parameter evaluated and passed to the parent CTOR.

Stuff like that is expected to be **covered** code, from what the community currently considers "coverage".

If you want to change that understanding, a new major release of `phpunit/php-code-coverage` would be needed, making such a change of intent clear (and then you should argue about it there).

The specific line at https://dashboard.stryker-mutator.io/reports/github.com/CuyZ/Valinor/master#mutant/Definition/Exception/ClassTypeAliasesDuplication.php is 100% to be considered as covered, if a test even touched that constructor.

mvorisek:
> @mvorisek the [line](https://github.com/CuyZ/Valinor/blob/6eba8109982a645ac9b2ff759e4f32cef65ba0e3/src/Definition/Exception/ClassTypeAliasesDuplication.php#L23) you highlighted is an integer **expression**, which is indeed a parameter evaluated and passed to the parent CTOR.

Expressionss in general are not coverd by this PR - like `'x';` (quite useless) or `public $x = 'x';` (property declaration), `static $x = 0;` (static variable probably not too).

> If you want to change that understanding, a new major release of `phpunit/php-code-coverage` would be needed, making such a change of intent clear (and then you should argue about it there).

in the history of this lib it was never a coverage line :), so I agree, this is a BC break that cannot land into 9.x if we follow semver strictly

sebastianbergmann:
I do not consider a change in what is / is not considered executable a BC break, to be honest.

Ocramius:
@sebastianbergmann that is fine, but then downstream hugely relies on it, at the moment :thinking:

sebastianbergmann:
Okay, then this should target `main` and the new major version that is due in February. Or can it be split into breaking (for `main`) and non-breaking (for `9.2`) changes?

Ocramius:
The idea is that this patch (#964) represents the state that consumers expect, and @mvorisek's ideas are probably for `master`.

That said, I've yet to do proper review on my end: @romm's testing provides the insight needed for now, which says that this patch works as "expected" :D

sebastianbergmann:
> The idea is that this patch (#964) represents the state that consumers expect, and @mvorisek's ideas are probably for `main`.

Thank you for clarifying this.

mvorisek:
> The idea is that this patch (#964) represents the state that consumers expect, and @mvorisek's ideas are probably for `main`.

no, I just clarified how the lib worked/works and where the current https://github.com/sebastianbergmann/php-code-coverage/issues/953 problem is

my question is still unanswered - what about transforming the current output using an optional analyser?

there will be no single BC break and will make everyone happy

@sebastianbergmann also when reviewing, please see my previous comments, the current implementation is not robust, it can mark lines as executed when they have been never executed/evaluated (in edge cases, mostly when code is badly formatted, but still, it can make developer lifes harder when they cannot rely 100% on the output data)

Ocramius:
> what about transforming the current output using an optional analyser?

Would mean downstream tooling is still broken, at this stage.

mvorisek:
> > what about transforming the current output using an optional analyser?
> 
> Would mean downstream tooling is still broken, at this stage.

please reread https://github.com/sebastianbergmann/php-code-coverage/pull/964#issuecomment-1348913202, the output needed for mutation analysis was never output by this lib

the breakage is also visible in the tests - they were dropped

when the newly proposed analyser would be made optional, mutation testing lib would have to be updated to enable the analyser to output the data needed (or convert any old/xdebug data using the new analyser)

Ocramius:
Feedback after trying this on my codebases is that this patch improves the coverage situation substantially.

Specifically, in https://github.com/Roave/BackwardCompatibilityCheck/pull/714, this led to results matching my expectations.

```
Metrics:
         Mutation Score Indicator (MSI): 94%
         Mutation Code Coverage: 94%
         Covered Code MSI: 100%
```

The remaining uncovered code is related to #945, otherwise this library would be at 100%, which I achieve locally with PCOV (for which I still need to adjust my CI docker images):

```
Metrics:
         Mutation Score Indicator (MSI): 100%
         Mutation Code Coverage: 100%
         Covered Code MSI: 100%
```

For comparison, the build fails when trying to upgrade `phpunit/php-code-coverage` in https://github.com/Roave/BackwardCompatibilityCheck/pull/707:

```
Metrics:
         Mutation Score Indicator (MSI): 89%
         Mutation Code Coverage: 89%
         Covered Code MSI: 100%
```

That code is 100% all executed and verified down to the last little corner.

Ocramius:
Excellent, thanks @sebastianbergmann! 

I sent a patch to re-enable `infection/infection` installation along with this package starting from `9.2.20`: https://github.com/infection/infection/pull/1777

## Regarding the disagreement on precision

@mvorisek as discussed above, constant expressions are "executed code" from our PoV.

I suggest you propose a change in direction with your own patch/proposal, making it clear that there's a breakage in underlying understanding of coverage, and accompanied with documentation of how to get the more "precise" results (according to your own understanding).

I am aware that @Slamdunk's patch leads to some edge cases in which precision is lost, but practically, the endgame for most people involved in this specific thread is mutation testing regardless, while **coverage on its own is** an entry point, but **nowadays an insufficient quality metric**.

From our PoV, getting the mutation testing tooling running is much more important than this precision, **at the moment**, so some false-positive coverage is worse than the false-negative coverage before this patch. Please be aware of this trade-off.

sebastianbergmann:
@Ocramius 9.2.21 has been tagged: https://github.com/sebastianbergmann/php-code-coverage/blob/9.2.21/ChangeLog.md#9221---2022-12-14

mvorisek:
@Ocramius I understand this PoV, but still I am not convinced it is good and I am not sure if anyone done review on the actual analyser impl. Fixing the precision will require complete revert of this PR and basically redone this PR from scratch :(

Ocramius:
@mvorisek and that's fine: I encourage you to go for it, if you have the time/energy for it :)

mvorisek:
Why I? I spent quite a lot of time on writing the original tests. They were dropped by author of this PR. Shouldn't be he responsible for taking care of them?

mvorisek:
@Slamdunk please fix https://github.com/atk4/ui/blob/dedb4cea3191e45ba2334ad48c0f98c6d25edf32/src/VueComponent/InlineEdit.php#L85

image

Slamdunk:
That's by design: that line is considered to be part of variable assignment now, not of the lambda.

Lambda content is correctly reported as non executed indeed.

mvorisek:
Closing bracket has a meaning. If reached, it means the function has reached the end. When you for example call `exit();` in it, it is not reached, and no coverage or red coverage should be shown. Rendering semicolon as covered is useless nor helpful with mutation testing.

Actually any software that honors the standard meaning, can look at the function as covered and it also returns https://github.com/sebastianbergmann/php-code-coverage/issues/954 bug to empty closures.

Slamdunk:
Empty methods and functions are handled as expected.
Empty closures can easily be cheated by declaring them as one-liner in any case, so I don't think we should put energies into handling them.
Well formatted closures are handled as expected, and I think the report you linked in https://github.com/atk4/ui/blob/dedb4cea3191e45ba2334ad48c0f98c6d25edf32/src/VueComponent/InlineEdit.php#L85 is good as is.

I'll be more than happy to fix any meaningful bug on closures, if you find one.

mvorisek:
I would be happy to hear feedback from @sebastianbergmann and @Ocramius too. I stand strongly by the fact that end bracket has meaning in line coverage.

Ocramius:
Please close the convo here and start a new one.

If you have a test case on which to discuss it, I suggest adjusting that and showing a failure as a start. The closure declaration seems correctly covered there.

Issue #974, opened 2022-12-23
Executable line analysis fails for declarations with enumerations and unions


davidbyoung:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 9.2.21
| PHP version               | 8.2.0
| Driver                    | Xdebug
| Xdebug version (if used)  | 3.2.0
| Installation Method       | Composer
| Usage Method              | PHPUnit
| PHPUnit version (if used) | 9.5.27

[Here is a scheduled run](https://github.com/aphiria/aphiria/actions/runs/3690919450/jobs/6248429526) with [100% code coverage](https://coveralls.io/builds/55136699) using 9.2.20.  [The next time it ran](https://github.com/aphiria/aphiria/actions/runs/3700259096/jobs/6268509356) with 9.2.21, [it reported a drop](https://coveralls.io/builds/55178228) due to a few interface methods and enum cases incorrectly being included in coverage.

sebastianbergmann:
CC @Slamdunk

sebastianbergmann:
This is likely caused by the changes discussed in #959 and #964.

Slamdunk:
Thanks for reporting this, I'll propose a fix in the next weeks

sebastianbergmann:
Thank you!

niconoe-:
I'm having this issue too so thank you very much @Slamdunk for fixing it very fast! Do you know if this could be released (for instance in a version 9.2.23) soon?

Thanks again!

Slamdunk:
I don't know, but you can use https://github.com/cweagans/composer-patches in the meantime

davidbyoung:
Happy new year! 9.2.23 did seem to resolve some of the issues, but I noticed that it's still trying to count interface methods with nullable union types containing enums. [Here is the latest Coveralls report with 9.2.23](https://coveralls.io/builds/55543365).

Slamdunk:
Happy new year to you too @davidbyoung https://github.com/sebastianbergmann/php-code-coverage/pull/977 😉

sebastianbergmann:
Can this be closed?

davidbyoung:
It looks like it can.  [My builds are back to 100%](https://coveralls.io/builds/55574731).  Thanks a bunch!

PR #976, opened 2022-12-24
Fix Enum and ComplexType

Slamdunk:
Fix https://github.com/sebastianbergmann/php-code-coverage/issues/974


sebastianbergmann:
Which celestial object are you on, Filippo? Here on earth only a couple of hours have passed, not weeks ;-)

Slamdunk:
Professional habit while talking to my stakeholder: only ever give the upper limit 😛

PR #985, opened 2023-03-03
Add description to suggested composer packages

AJenbo:
The value in suggestions is not for the version (like in required) but for a descriptive test that can help developers in deciding if they should install the suggested package.

This hint will be show to developers when running `composer suggests`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,8 +47,8 @@
         "phpunit/phpunit": "^10.1"
     },
     "suggest": {
-        "ext-pcov": "*",
-        "ext-xdebug": "*"
+        "ext-pcov": "PHP extension that provides line coverage",
+        "ext-xdebug": "PHP extension that provides line coverage as well as branch and path coverage"
     },
     "autoload": {
         "classmap": [


composer.json
```diff
@@ -47,8 +47,8 @@
         "phpunit/phpunit": "^10.1"
     },
     "suggest": {
-        "ext-pcov": "*",
-        "ext-xdebug": "*"
+        "ext-pcov": "A fast code coverage driver",
```
sebastianbergmann, line 50: A better description would be "PHP extension that provides line coverage".

composer.json
```diff
@@ -47,8 +47,8 @@
...
+        "ext-xdebug": "Code coverage driver"
```
sebastianbergmann, line 51: A better description would be "PHP extension that provides line coverage as well as branch and path coverage".

AJenbo: Thanks, I updated both descriptions

sebastianbergmann:
Cherry-picked into `9.2` and merged to `main` from there.

AJenbo:
Thank you for the quick review :)

PR #1005, opened 2023-07-14
Add cache to Coverage Report Directory

ruifil:
Add cache to Coverage Report Directory 

- To avoid issues with multiple reporter when using PHP fibers when generate a second report
- Optimize the speed generation of next reports


sebastianbergmann:
Merged manually, thanks.

Issue #1009, opened 2023-09-12
Coverage Report cache makes PHP report memory usage explode


Slamdunk:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 10.1.3
| PHP version               | 8.2
| PHPUnit version (if used) | 10.3.3

Changes merged in https://github.com/sebastianbergmann/php-code-coverage/pull/1005 and released [`10.1.3`](https://github.com/sebastianbergmann/php-code-coverage/compare/10.1.2...10.1.3) made `Report\PHP` unusable as now it always consumes too much memory.

Our `coverage.php` file went from `29 Mb` to `737 Mb`, re-processing it with `phpcov` or any other tool skyrockets the memory used from `< 1 Gb` to `> 4 Gb`, with no control over it and, most important, no real benefit for the purposes of re-processing many times the `Report\PHP` file.

https://github.com/sebastianbergmann/php-code-coverage/pull/1005 is good because when you use many `Report\*`,  the `Directory` is build only once and PHPUnit's report generation is faster, but `Report\PHP` should serialize the `CodeCoverage` object **without** the `$this->cachedReport` property filled.

I suggest the following improvements:

- [x] `Report\PHP` should serialize the `CodeCoverage` object with the cache cleared using a new `@internal` API such as `CodeCoverage::clearCache` (different from the currently present `CodeCoverage::clear` which has another purpose)
    * https://github.com/sebastianbergmann/php-code-coverage/pull/1011
- [x] [`PHPUnit\Runner\CodeCoverage`](https://github.com/sebastianbergmann/phpunit/blob/10.3.3/src/Runner/CodeCoverage.php) should generate the `Report\PHP` as first in the list, so that when it's processed no cache has been already built; the aforementioned API `CodeCoverage::clearCache` would be practically useless, but it's still needed to ensure consistency of `php-code-coverage` package
    * https://github.com/sebastianbergmann/phpunit/pull/5508
    * https://github.com/sebastianbergmann/phpunit/pull/5509
- [x] [`SebastianBergmann\PHPCOV\Command`](https://github.com/sebastianbergmann/phpcov/blob/9.0.1/src/cli/Command.php) should generate the `Report\PHP` as first in the list for the same reason
    * https://github.com/sebastianbergmann/phpcov/pull/134

sebastianbergmann:
All of your suggestions make sense to me, at least as short-term solutions.

Once #874 has been implemented, we should have a long-term solution.

PR #1011, opened 2023-09-12
Avoid serialization of cache data in PHP report

Slamdunk:
Fix https://github.com/sebastianbergmann/php-code-coverage/issues/1009


Slamdunk:
This PR should be easily mergeable into upstream branch `main`, so no need for a separate PR as far as I can tell

Issue #1012, opened 2023-09-15
Cobertura report pulls functions from report scope, not the individual element


SolovievEvgenii:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 9.2.28
| PHP version               | 7.4
| Driver                    | Xdebug
| Xdebug version (if used)  | 3.1.6
| Installation Method       | Composer
| Usage Method              | PHPUnit
| PHPUnit version (if used) | 9.6.12

We are working with one legacy PHP project that contains about 3000 classes and over 1000 global functions.
We try to plug in the code coverage feature to our gitlab.
When I tried to export the coverage report to Cobertura XML format, I received a Killed Error from Linux:

```text
[19369437.006651] Out of memory: Kill process 27901 (php) score 680 or sacrifice child
[19369437.007441] Killed process 27901 (php) total-vm:22660172kB, anon-rss:22550668kB, file-rss:0kB, shmem-rss:0kB
[19369437.554951] oom_reaper: reaped process 27901 (php), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB
```

I simulated the situation by simple files (https://github.com/SolovievEvgenii/cobertura_bug1), debugged the [Cobertura class](https://github.com/sebastianbergmann/php-code-coverage/blob/main/src/Report/Cobertura.php) and found that the nodes of all our global functions is included in the body of each package (which corresponds to a single file), although in reality the file may not have any relationship with global functions, no calls. no definitions. Look at the screenshot with resulted XML and the corresponded codes:

![изображение](https://github.com/sebastianbergmann/php-code-coverage/assets/145051525/d2070c89-7d5a-4dfe-ae64-03b9ff448df1)

Pay attention at nodes, underlined by magenta lines, these nodes are talking about two different methods, but both are pointing to the same line. It is wrong, is it? It happens because at each package's iteration the list of functions is taken from global scope instead of list from package's functions

To get the fixed XML I added to your code my little crutch, I take a list of functions not from an iterating element "item", but from element from "children" array of the "report" object.

See the function with my try to fix [here](https://github.com/SolovievEvgenii/cobertura_bug1/blob/main/src/Helpers.php#L26)

At last I'he got the cobertura format of our big project without any killings. But I am not sure that it is really right, is it difficult to validate the report of large project. I suspect that my fix is not optimized way and hope to receive feedback about the correctness of this approach.

sebastianbergmann:
[...]

PR #1013, opened 2023-09-15
#1012 Cobertura report pulls functions from report scope, not the individual element

dvdoug:
Haven't tested this myself yet, but [SolovievEvgenii](https://github.com/SolovievEvgenii) could you test to see if it works for you?

### DIFF
--- a/src/Report/Cobertura.php
+++ b/src/Report/Cobertura.php
@@ -194,7 +194,7 @@ final class Cobertura
                 }
             }
 
-            if ($report->numberOfFunctions() === 0) {
+            if ($item->numberOfFunctions() === 0) {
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Thank you, @dvdoug!

@SolovievEvgenii Can you test this, please?

bastien-phi:
Good job @dvdoug ! I use cobertura reports every day and never noticed that it had issues with functions (I honestly never read them). Thanks for the fix.

After testing these changes, the report looks way better. The coverage of the functions is not duplicated anymore.

@sebastianbergmann You can wait @SolovievEvgenii to reply if you want but it looks good to me !

sebastianbergmann:
@bastien-phi Thank you!

PR #1016, opened 2023-10-20
[CSS] Changed the x direction of scrollbox from hidden to scroll

SakiTakamachi:
Hi

When checking project risks on the dashboard, if the class name is too long, the table will be cut off and CRUP cannot be checked.

It seemed like it would be a good idea to simply change hidden to scroll using CSS, so I would like to make such a suggestion. Is the target branch 9.2 ok?

Best regards.

cc: @sebastianbergmann


sebastianbergmann:
Thank you!

PR #1019, opened 2023-11-15
Fix interface precedence over method handling

Slamdunk:
Fix https://github.com/sebastianbergmann/php-code-coverage/issues/1018


sebastianbergmann:
Thanks!

Issue #1020, opened 2023-11-20
Single line method is ignored


mvorisek:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 9.2.29
| PHP version               | 8.2.12
| Driver                    | Xdebug
| PCOV version (if used)    | n/a
| Xdebug version (if used)  | 3.2.2
| Installation Method       | Composer
| Usage Method              | PHPUnit
| PHPUnit version (if used) | 9.6.13

repro code:
```
class InitializerTraitTest extends TestCase
{
    public function testInitDeclaredPublicException(): void
    {
        $m = new class() extends AbstractInitializerMock {
            public function init(): void {}
        };
    }
}
```

(full source code: https://github.com/atk4/core/blob/5.0.0/tests/InitializerTraitTest.php#L72)

when the code above is modified like:

```diff
+            #[\Override]
             public function init(): void {}
```

then the coverage is collected. It seems this lib does not account for single line methods (methods starting and ending at the same line, even a method attribute on another line is enough to enable coverage).

![image](https://github.com/sebastianbergmann/php-code-coverage/assets/2228672/d64ee165-2b8c-4ff1-a566-b1cfb557533e)

sebastianbergmann:
CC @Slamdunk

Slamdunk:
Looking at here:

https://github.com/atk4/core/blob/9bc701001c6c7959d12b1a66cdeea5d618735abb/src/InitializerTrait.php#L46-L50

It seems to me that the CC is correctly corrected both times:

1. when the method is `public`, `init()` is **not** called so it's red
2. when the method is `protected`, `init()` **is** called so it's green

The `#[\Override]` attribute has nothing to do with this.

I see no bug here :shrug:

sebastianbergmann:
@Slamdunk Thank you!

Slamdunk:
@mvorisek now it makes mor sense: fix proposed in https://github.com/sebastianbergmann/php-code-coverage/pull/1021

PR #1021, opened 2023-11-21
Fix one-line method for anonymous classes

Slamdunk:
Fix https://github.com/sebastianbergmann/php-code-coverage/issues/1020

### DIFF
--- a/tests/_files/source_for_branched_exec_lines.php
+++ b/tests/_files/source_for_branched_exec_lines.php
@@ -460,12 +460,18 @@ class MyClass
                 {
                     return;                 // +3
                 }
-            }                               // -3
+
+                public function m1(): void {} // +1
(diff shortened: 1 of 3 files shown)


tests/_files/source_for_branched_exec_lines.php
```diff
@@ -460,12 +460,18 @@ public function myMethod()
                 {
                     return;                 // +3
                 }
-            }                               // -3
+
+                public function m1(): void {} // +1
```
mvorisek, line 464: please add a test for a method in non-anonymous class as well (at least in this test file it is not tested currently)

Slamdunk: Done :+1:

mvorisek:
Thank you both ❤️

PR #1028, opened 2024-01-25
Implement processed code coverage data mapper

AndrewFeeney:
This is my first pass at implementation for an object `SebastianBergmann\CodeCoverage\Data\ProcessedCodeCoverageDataMapper` which serializes to JSON and de-serializes from JSON a given instance of `SebastianBergmann\CodeCoverage\Data\ProcessedCodeCoverage`.

The intention here is to move in the direction of being able to generate code coverage reports that do not require the serialize() and unserialize() PHP functions, and instead rely on a data only JSON format which might be less prone to issues such as https://github.com/sebastianbergmann/phpcov/issues/109. This PR attempts to implement a basic JSON format only, and does not fully realize that goal by replacing the current implementation of serialization or documentation.

The JSON format have essentially represents the underlying data in the class. i.e.

```javascript
{
   "lineCoverage": [
       // ...
   ],
   "functionCoverage": [
       // ...      
   ],
}
```

I have not attempted to include any of the meta data that might be used to instantiate a `Driver` or `Filter` instance which are required to instantiate a `CodeCoverage` instance.

* My Hesitations About My Implementation

I'm not sure if it would be better to make the JSON format more general, and nest the above properties under something like `coverage` using other keys, perhaps `driver` and `filter` or something else to include other meta data about the code coverage report. I have omitted this based on your stated preference to de-serialize to `ProcessedCodeCoverageReport`. I would welcome your guidance here.
   
 * My Hesitations About My Tests
 
I'm still familiarizing myself with the codebase and the shape of the coverage data, so I'm not certain I've covered everything I need to in testing.
 
I've included one "End to End" style test for XML which may not be appropriate to merge. This was more about me ensuring that I had everything I needed to replicate a known expected XML report. I'm not sure how far to go down this path in adding more tests like that, as it may be redundant, and I don't want to be retesting the XML report generation implementation. Once again, I'd welcome your opinion here.

### DIFF
--- a/src/Data/ProcessedCodeCoverageDataMapper.php
+++ b/src/Data/ProcessedCodeCoverageDataMapper.php
@@ -18,10 +18,12 @@ namespace SebastianBergmann\CodeCoverage\Data;
  */
 final class ProcessedCodeCoverageDataMapper
 {
+    const KEY_LINE_COVERAGE = 'line_coverage';
+
     public function toJson(ProcessedCodeCoverageData $processedCodeCoverageData): string
     {
         $arrayMapping = [
-            'line_coverage' => $processedCodeCoverageData->lineCoverage(),
+            self::KEY_LINE_COVERAGE => $processedCodeCoverageData->lineCoverage(),
         ];
 
         return json_encode($arrayMapping);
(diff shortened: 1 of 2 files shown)


AndrewFeeney:
I see that there are some CI failures which I'll take a look at next. I also should mention that my implementation here is laughably simple. One thing I haven't looked at is validating the JSON on unserialization. I'm not sure if that's necessary or not, but I can have a go at that if you'd like.

sebastianbergmann:

Issue #1038, opened 2024-06-29
CI pipeline for `9.2` is broken


sebastianbergmann:
Before c00148658dc981cc32ae6af1f861f6aee857fc1f, the `composer.json` file of this project configured a dependency on `"phpunit/phpunit": "^9.3"`.

I noticed that on GitHub Actions, Composer would install PHPUnit 9.4.4 instead of PHPUnit 9.6.19 (which is the latest version that fulfills the `^9.3` version constraint. See https://github.com/sebastianbergmann/php-code-coverage/actions/runs/9722443548/job/26836345492#step:6:27.

I changed this version constraint from `^9.3` to `^9.6` in c00148658dc981cc32ae6af1f861f6aee857fc1f in order to force an error on GitHub Actions when Composer installs the dependencies and provides an error message that explains why PHPUnit 9.4.4 was installed before:

```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - phpunit/phpunit[9.6.0, ..., 9.6.11] require phpunit/php-code-coverage ^9.2.13 -> satisfiable by phpunit/php-code-coverage[9.2.13, ..., 9.2.x-dev] from composer repo (https://repo.packagist.org) but phpunit/php-code-coverage is the root package and cannot be modified. See https://getcomposer.org/dep-on-root for details and assistance.
    - phpunit/phpunit[9.6.12, ..., 9.6.19] require phpunit/php-code-coverage ^9.2.28 -> satisfiable by phpunit/php-code-coverage[9.2.28, ..., 9.2.x-dev] from composer repo (https://repo.packagist.org) but phpunit/php-code-coverage is the root package and cannot be modified. See https://getcomposer.org/dep-on-root for details and assistance.
    - Root composer.json requires phpunit/phpunit ^9.6 -> satisfiable by phpunit/phpunit[9.6.0, ..., 9.6.19].
```

[`COMPOSER_ROOT_VERSION` is set to `9.2-dev`](https://github.com/sebastianbergmann/php-code-coverage/blob/39d628812d8d83344a6c1b07799e3700d830d355/.github/workflows/ci.yml#L10) in the GitHub Actions configuration and a [`branch-alias`](https://github.com/sebastianbergmann/php-code-coverage/blob/39d628812d8d83344a6c1b07799e3700d830d355/composer.json#L66) is configured in `composer.json`.

[...]

PR #1040, opened 2024-07-03
Fix: Adjust `COMPOSER_ROOT_VERSION` and branch alias

localheinz:
This pull request

- [x] removes the declaration of the `COMPOSER_ROOT_VERSION` environment variable

Related to #1038.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,9 +6,6 @@ on:
 
 name: CI
 
-env:
-  COMPOSER_ROOT_VERSION: "9.2-dev"
-
 jobs:
   coding-guidelines:
     name: Coding Guidelines
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
 
 name: CI
 
+env:
+  COMPOSER_ROOT_VERSION: "9.2.x-dev"
+
 jobs:
   coding-guidelines:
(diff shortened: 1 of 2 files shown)


localheinz:
Hmm.

localheinz:
@sebastianbergmann 

Running

```php
<?php

declare(strict_types=1);

use Composer\Semver;

require_once  __DIR__ . '/../vendor/autoload.php';

$versions = [
    '9.2-dev',
    '9.2.x-dev',
    '9.3-dev',
    '9.3.x-dev',
];

$constraint = '^9.2.13';

var_dump(array_combine(
    $versions,
    array_map(static function (string $version) use ($constraint): string {
        return sprintf(
            '%s %s',
            Semver\Semver::satisfies($version, $constraint) ? 'satisfies' : 'does not satisfy',
            $constraint
        );
    }, $versions)
));
```

yields

```
array(4) {
  ["9.2-dev"]=>
  string(24) "does not satisfy ^9.2.13"
  ["9.2.x-dev"]=>
  string(17) "satisfies ^9.2.13"
  ["9.3-dev"]=>
  string(17) "satisfies ^9.2.13"
  ["9.3.x-dev"]=>
  string(17) "satisfies ^9.2.13"
}
```

Does that make sense?

sebastianbergmann:
> Does that make sense?

No, but it seems to fix the build :-) Thanks!

PR #1037, opened 2024-05-11
Bootstrap 5 upgrade without theme feature.

gammamatrix:
## Changes
- The theme feature was removed in this branch.
- See GH-1036

PR #1046, opened 2024-09-05
GH-1037 - CSS fixes for first column of code review table.

gammamatrix:
## Changes
- I did a quick test and it appears to work fine in Chrome and Firefox, but the selector does not appear to work on old Safari.

### DIFF
--- a/src/Report/Html/Renderer/Template/css/style.css
+++ b/src/Report/Html/Renderer/Template/css/style.css
@@ -147,14 +147,13 @@ table + .structure-heading {
  padding-top: 0.5em;
 }
 
-table#code {
-  td:nth-child(1) {
-    padding-left: .75em;
-    padding-right: .75em;
-    a {
-      text-decoration: none;
-     }
-  }
+table#code td:first-of-type {
+  padding-left: .75em;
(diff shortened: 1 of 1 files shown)


gammamatrix:
## Changes

```css
table#code td:first-of-type {
  padding-left: .75em;
  padding-right: .75em;
}

table#code td:first-of-type a {
  text-decoration: none;
}
```
- This appears to be working on Chrome, Firefox and Safari.
- Did I cover the padding in the right spot?

sebastianbergmann:
I still see the whitespace after this PR on Firefox 130 and Chrome 128:

grafik

gammamatrix:
Is that extra space just from the flex box and making the view port over 2000px?

I did a comparison... I marked the images, the two in the middle are for Firefox 130 with Bootstrap 5 on the left and Bootstrap 4 on the right.
- The outside browsers are both Chrome:

GH-1046-chrome-128-firefox-130-with-bootstrap-4-and-5

sebastianbergmann:
> Is that extra space just from the flex box and making the view port over 2000px?

Maybe. I always have my browser windows full-screen :)

PR #1051, opened 2024-12-09
Fix `phpstan-type` declarations

staabm:
[the syntax does not have a `=`](https://phpstan.org/writing-php-code/phpdoc-types#local-type-aliases)


staabm:
now tested locally. it does not fix the issue, but still is a required change IMO

sebastianbergmann:
Thank you!

PR #1054, opened 2024-12-11
Switch to click event for toggling coverage popover (closes #1053).

### LINKED ISSUE #1053: Usability: Change hover to click in html report
| php-code-coverage version | 11.0.z

I like the html report a lot, but it is not usable for us with javascript enabled. The popover displaying the covering lines is unreadable because it is so huge and is not scrollable. And also scrolling through the page itself with the mouse inside is very annoying due to the constant popup blocking the view.

I would like to ask if we could change the `mouseenter`/`mouseleave`  events in `src/Report/Html/Renderer/Template/js/file.js` to toggle with a `click` event instead. Then I hope the popover will stay open, allows to scroll it, and doesn't come in the way when browsing a page.

I would create a PR if this was acceptable.

uncaught:


sebastianbergmann:
Cherry-picked into `11.0` and merged to `main` from there. Thank you!

Issue #1075, opened 2025-04-27
Dashboard chart of method coverage fails on mouse hover


mytskine:
When moving the mouse over a point of the scatter chart of methods, nothing happens with Chromium 134. For each hovering, the JavaScript console displays a unhelpful error deep into nv.d3.min.js. The behavior is the same with Firefox ESR 128.9.

At a time, both scatter charts were broken, as reported in #440. The first chart (classes) was fixed but the second one (methods) was not, as someone reported in the same issue (after it was closed).

Since the class and method charts are similar, I supposed the browser, the bug was in the library used to build the SVG charts. Unfortunately, [nvd3](https://github.com/novus/nvd3)'s last release was in 2017. It's a layer above d3js, but no longer compatible with recent releases of d3js. I'll submit a merge request that replaces nvd3 with a maintained library.

sebastianbergmann:
> I'll submit a merge request that replaces nvd3 with a maintained library.

Thank you!

PR #1076, opened 2025-04-27
report dashboard uses lib billboard.js for charts (replaces nvd3, unmaintained)

mytskine:
Issue #1075

I quickly reviewed the charting libraries based on d3js that could replace nvd3. In my opinion, the only sane choice was [billboard.js](https://naver.github.io/billboard.js/).

I tried to minimize the changes, visually and in the PHP code, but 2 visual changes were made on purpose:
- The horizontal and vertical ticks from the scatter graphs were removed. With a high-density of near values, it created continuous thick segments which were meaningless.
- The tooltip is no longer a link, but each dot, when clicked, will send to the related page. This was a bug, because the link could not be reached, as the tooltip disappeared as soon as the mouse left the point.

The JS code is plain, though a bit more modern: some ES6 features, and no more `var`. It should work in any browser from the last decade.


sebastianbergmann:
Thank you!

PR #1087, opened 2025-08-28
Avoid extra loop

sebastianbergmann:


sebastianbergmann:
@staabm Could I bother you to take a look? The tests pass, but yet I am not entirely sure this is better.

staabm (commented):
lgtm - I have no proof though whether this is faster or not.

looks like it leads to less work when classes don't have a parent class

PR #1088, opened 2025-08-28
run array_unique() only once per class

staabm:
at this point its mostly guess work.

having loops after each other should be faster then nested loops though - O(3n) vs. O(n^2).


staabm:
we need to be more incremental as our problem is memory, not cpu: https://github.com/sebastianbergmann/php-code-coverage/pull/1089

PR #1105, opened 2025-11-08
Represent CoverageData in objects

staabm:
thesis is, that we need less memory when processing big projects

in phpstan-src I can see the coverage producing test-runs to consume ~3GB of data.
my gut feeling is, that if we would consume less data we would automatically get faster, because we need less memory allocations

---

running phpstan-src@8f9490ecc6

before this PR (https://github.com/sebastianbergmann/php-code-coverage/commit/37047b5f0e98b31dccb9b7e7241e4f66ebfcd358):

```
➜  phpstan-src git:(2.1.x) ✗ XDEBUG_MODE=coverage php vendor/bin/phpunit --coverage-xml=tmp/coverage-xml tests/PHPStan/Type/Constant --path-coverage
PHPUnit 12.4.2 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.27 with Xdebug 3.4.7
Configuration: /Users/staabm/workspace/phpstan-src/phpunit.xml

...............................................................  63 / 128 ( 49%)
............................................................... 126 / 128 ( 98%)
..                                                              128 / 128 (100%)

Time: 01:18.247, Memory: 1.18 GB

OK (128 tests, 209 assertions)

Generating code coverage report in PHPUnit XML format ... done [00:05.877]
```

with this PR (https://github.com/sebastianbergmann/php-code-coverage/pull/1105/commits/1b73b3ca0d35a186ae127b374381fbd3089690f3)

```
➜  phpstan-src git:(2.1.x) ✗ XDEBUG_MODE=coverage php vendor/bin/phpunit --coverage-xml=tmp/coverage-xml tests/PHPStan/Type/Constant --path-coverage
PHPUnit 12.4.2 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.27 with Xdebug 3.4.7
Configuration: /Users/staabm/workspace/phpstan-src/phpunit.xml

...............................................................  63 / 128 ( 49%)
............................................................... 126 / 128 ( 98%)
..                                                              128 / 128 (100%)

Time: 00:53.183, Memory: 1.18 GB

OK (128 tests, 209 assertions)

Generating code coverage report in PHPUnit XML format ... done [00:05.709]
```

PR #1109, opened 2025-11-10
Represent line-coverage data in objects

staabm:
analog https://github.com/sebastianbergmann/php-code-coverage/pull/1105 but for regular line coverage data

### DIFF
--- a/src/Report/Xml/Facade.php
+++ b/src/Report/Xml/Facade.php
@@ -177,23 +177,23 @@ final class Facade
 
     private function processUnit(ProcessedClassType|ProcessedTraitType $unit, Report $report): void
     {
-        if (isset($unit['className'])) {
+        if ($unit instanceof ProcessedClassType) {
             $unitObject = $report->classObject(
-                $unit['className'],
-                $unit['namespace'],
-                $unit['startLine'],
-                $unit['executableLines'],
-                $unit['executedLines'],
-                (float) $unit['crap'],
+                $unit->className,
+                $unit->namespace,
+                $unit->startLine,
+                $unit->executableLines,
+                $unit->executedLines,
+                (float) $unit->crap,
             );
         } else {
             $unitObject = $report->traitObject(
-                $unit['traitName'],
-                $unit['namespace'],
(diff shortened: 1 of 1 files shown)


staabm:
very interessting.. this PR does not yet yield a measurable performance improvement in the cases I tested. neither perf nor memory wise.

still its a improvement regarding maintainability and readability.

sebastianbergmann:
Why is this still a draft? Thanks.

staabm:
because it did not yield the expected perf results. if you are fine with the change we can still go ahead

staabm:
I think this is good to go.

src/Data/ProcessedClassType.php
```diff
@@ -0,0 +1,37 @@
...
+ */
+namespace SebastianBergmann\CodeCoverage\Data;
+
+/**
+ * @internal This class is not covered by the backward compatibility promise for phpunit/php-code-coverage
+ */
+final class ProcessedClassType
+{
+    public function __construct(
+        public readonly string $className,
```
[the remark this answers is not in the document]
staabm: here we go.

remaining CI error is related to a deprecated php-cs-fixer setting, not the PR itself

src/Node/AbstractNode.php
```diff
@@ -186,11 +185,11 @@ public function cyclomaticComplexity(): int
         $ccn = 0;
 
         foreach ($this->classesAndTraits() as $classLike) {
-            $ccn += $classLike['ccn'];
+            $ccn += $classLike->ccn;
```
sebastianbergmann: Is there really a measurable performance benefit in using a public property over a method call to a public method that returns the value of a private property?

staabm: Yes, doing so in https://github.com/sebastianbergmann/php-code-coverage/pull/1105

staabm:
Btw: motivation for this PR is also more strict structures, so we can level up phpstan and prevent bugs in the future

staabm:
would be great to get this merged, as it likely conflicts with other de-DOM-mification work along the way

PR #1110, opened 2025-11-10
Use plain `for`-loop over `foreach(range() ..)`

staabm:
`range()` is showing up in the top 5 memory consumers when running 

`XDEBUG_MODE=coverage blackfire run php vendor/bin/phpunit --coverage-xml=tmp/coverage-xml tests/PHPStan/Rules/Api/`

on phpstan-src

grafik

---

PS: `token_get_all` will be more efficient with https://github.com/theseer/tokenizer/pull/35

### DIFF
--- a/src/Node/File.php
+++ b/src/Node/File.php
@@ -11,7 +11,6 @@ namespace SebastianBergmann\CodeCoverage\Node;
 
 use function array_filter;
 use function count;
-use function range;
 use SebastianBergmann\CodeCoverage\CodeCoverage;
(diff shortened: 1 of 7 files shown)


sebastianbergmann:
Why is this still a draft? Thanks.

staabm:
I realized the actual root cause of `range` showing up in the perf profile, is the occurence in this spot:

https://github.com/sebastianbergmann/php-code-coverage/blob/37047b5f0e98b31dccb9b7e7241e4f66ebfcd358/src/Target/MapBuilder.php#L250

and I did not yet find a alternative way to implement it. the changed source in this PR until now would not affect performance

PR #1112, opened 2025-11-20
Project: Make parent::__construct() call explicit

staabm:


src/Report/Xml/Project.php
```diff
@@ -79,19 +88,6 @@ public function asDom(): DOMDocument
         return $this->dom();
     }
 
-    private function init(): void
-    {
-        $dom = new DOMDocument;
-        $dom->loadXML('<?xml version="1.0" ?><phpunit xmlns="https://schema.phpunit.de/coverage/1.0"><build/><project/></phpunit>');
-
-        $this->setContextNode(
```
staabm: this call was equivalent to the baseclass `__construct()` implementation.. but we shouldn't rely on implementation knowledge across classes, when not necessary

PR #1119, opened 2025-11-20
Remove getElementsByTagNameNS() for cases which can never find a element

staabm:


src/Report/Xml/Project.php
```diff
@@ -63,19 +63,12 @@ public function buildInformation(): BuildInformation
 
     public function tests(): Tests
     {
-        $testsNode = $this->contextNode()->getElementsByTagNameNS(
```
staabm: we call this method on a document, which at this point of time cannot already contain a `<tests>` element.

Issue #1131, opened 2025-12-19
Invalid XML generated when both PCOV and Xdebug are loaded


stijnh92:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 12.5.1
| PHP version               | 8.3.28
| Driver                    | PCOV / Xdebug
| PCOV version (if used)    | 1.0.12
| Xdebug version (if used)  | 3.5.0
| Installation Method       | Composer
| Usage Method              | PHPUnit
| PHPUnit version (if used) | 12.5.4

I have created a script that loads the XML coverage file, generated by phpunit, to publish some statistics of our codebase. Since the latest version of php-code-coverage, I am unable to load the XML file with the following error:
```
PHP Warning:  simplexml_load_file(): index.xml:5: parser error : Attribute name redefined
PHP Warning:  simplexml_load_file():     <driver name="xdebug" version="3.5.0" name="pcov" version="1.0.12"/>
```

Looking at the XML file, I can see that the `name` attribute is used twice on the `driver` tag which is invalid XML:
```xml
<driver name="xdebug" version="3.5.0" name="pcov" version="1.0.12"/>
```

We use PCOV in our pipelines where the phpunit job is executed using a PHP image that has Xdebug enabled by default. We don't really need Xdebug so we can use another image to work around this issue, but I just wanted to report that when using a combination of Xdebug and PCOV, invalid XML is being generated which can not be loaded by the `simplexml_load_file` function.

staabm:
@stijnh92 thanks for the report. please give the fix in https://github.com/sebastianbergmann/php-code-coverage/pull/1132 a try

stijnh92:
@staabm perfect, that works 🙏

PR #1132, opened 2025-12-19
Coverage-xml cannot contain pcov and xdebug at the same time

staabm:
attribute name needs to be unique

closes https://github.com/sebastianbergmann/php-code-coverage/issues/1131

---

this regressed in https://github.com/sebastianbergmann/php-code-coverage/commit/5b74f6242eb0baf4414dd15033afd51abc0bb45d which previously used DOM to de-duplicate the attribute name automatically because it just overwrote the previous value


stijnh92:
Can confirm that this works and generates valid XML.

sebastianbergmann:

PR #1141, opened 2026-03-14
Improve visualization of branch coverage and path coverage in the HTML report

sebastianbergmann:
When code coverage data is collected with Xdebug's path coverage mode, the HTML report generates three views for each source file: line coverage, branch coverage, and path coverage. These views have been significantly improved.

## Navigation between views

Each file page now has tabs for "Line Coverage", "Branch Coverage", and "Path Coverage" at the top.

Previously, the branch coverage and path coverage pages were generated but not linked from anywhere; you had to know that they exist and manually edit the URL to reach them. Now you can switch between the three views with a single click, and the active view is highlighted.

### Screenshots

grafik

## Branch coverage view

The source code listing now has a gutter column with decision point markers: on lines where the control flow branches (for instance an `if` statement or a ternary expression), one dot is shown per possible outcome, in the report's "covered" color when that outcome was executed by a test and in the "not covered" color when it was not. This makes it easy to spot, for example, an `if` statement whose `else` case is never exercised, even when the line itself is shown as (partially) covered.

The "Branches" section below the source code has been redesigned. Previously, the source code lines of every branch were repeated there, which made the section long and repetitive. Each method now gets a compact table with one row per branch, showing

* the lines the branch spans (linked to the source code listing above),
* whether the branch is covered, and
* which tests cover it (shown in a popover, like in the source code listing).

Each method heading now also shows a summary badge such as `3/4`, the number of covered branches out of the total number of branches, using the report's colors for fully covered, partially covered, and not covered code.

### Screenshots

grafik

## Path coverage view

The "Paths" section has been redesigned in the same way. Each method gets a table with one row per path through the method, showing the sequence of branches that make up the path (as links to the source code lines), whether the path is covered, and which tests cover it. A summary badge on the method heading shows the number of covered paths out of the total number of paths.

New in this view: a control flow graph for each method, rendered as an image. The nodes of the graph are the method's branches (labelled with the source code lines they span), connected by arrows that show how control can flow from one branch to another, from entry to exit. Covered branches and transitions are drawn in the report's "covered" color, uncovered ones in the "not covered" color. The graph uses the same color scheme as the rest of the report, including a custom color scheme, when one is configured, and adapts to light mode and dark mode.

The table and the graph are connected: clicking on a row in the paths table highlights the corresponding path in the control flow graph, so you can see exactly which route through the method a path represents. Clicking the row again removes the highlight.

Rendering the control flow graph requires the `dot` tool from [Graphviz](https://graphviz.org/) to be available on the system that generates the report. When `dot` is not available, the report is generated without control flow graphs.

Methods with more than 100 paths were previously not listed at all; only a note suggesting a refactoring was shown. For such methods, the first 100 paths are now rendered inside a collapsed section that can be expanded on demand, together with a note stating how many paths there are in total. The control flow graph of such a method still reflects the coverage of all its branches; only the interactive highlighting is limited to the listed paths.

### Screenshots

grafik
grafik
grafik
grafik


sebastianbergmann:
Both #1140 and this have the same [ancestor](https://github.com/sebastianbergmann/php-code-coverage/commit/19a9bf4115710b489544fefc4cf3090a389f21d9), but #1140 is the bigger change and should be merged first, in case both #1140 and this will be merged.

Issue #1148, opened 2026-04-09
Coverage regression with v14.0.0 and Paratest


baggids:
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 14.0.0
| PHP version               | 8.5.4
| Driver                    | Xdebug
| Xdebug version (if used)  | 8.5.1
| Installation Method       | Composer
| Usage Method              | Paratest v7.22.2
| PHPUnit version (if used) | 13.1.1

We have a Symfony + ApiPlatform stack. After bumping php-code-coverage (and related dependencies) coverage % dropped for PHP entity files covered by unit and functional tests. The no. of lines covered remained the same but the denominator went up to include multi-line PHP Attributes. For example, the following lines are reported as uncovered whereas with php-code-coverage v13.0.1 they were neither covered/uncovered.
```
#[ApiResource(
    operations: [
        new Get(
            openapi: new Operation(summary: ....

#[Groups([
     'read:output'
])]
```
Strangely, running paratest with --processes=1 (instead of 16) restored the coverage % and the above lines were not reported as uncovered and the denominator lines matched the coverage from before the version bump.

sebastianbergmann:
Nothing changed in php-code-coverage 14.0.0 in regard to how a line of code is decided to be covered or not covered.

The fact that running ParaTest with `--processes=1` restores the previous behavior tells me that this is most likely related to how serializing, merging, and deserializing code coverage data now works in php-code-coverage 14.0.0 and how ParaTest interacts with that.

I do not know whether https://github.com/paratestphp/paratest/pull/1096 (which was merged yesterday) already solves your problem. It would be worth trying.

For this issue to be actionable for me, I need a minimum viable, self-contained example that demonstrates "something is wrong" in php-code-coverage when it is not used together with ParaTest.

CC @Slamdunk

Slamdunk:
@baggids I am unable to reproduce the bug, can you kindly provide a minimum viable, self-contained example?

baggids:
@sebastianbergmann @Slamdunk would this suffice?
https://github.com/baggids/php-code-coverage_regression

Slamdunk:
Perfect, thank you @baggids 

@sebastianbergmann you can close this as it's a ParaTest bug 🪳

baggids:
@Slamdunk should I link to this issue in [Paratest](https://github.com/paratestphp/paratest)?

Slamdunk:
[...]

PR #1149, opened 2026-04-12
Do not treat lines spanned by attributes as executable

Slamdunk:
Fix https://github.com/sebastianbergmann/php-code-coverage/issues/1148

Beware the targeted branch is `12.5`, so releases are needed also for `13` and `14` branches.

### DIFF
--- a/src/StaticAnalysis/Visitor/ExecutableLinesFindingVisitor.php
+++ b/src/StaticAnalysis/Visitor/ExecutableLinesFindingVisitor.php
@@ -83,7 +83,9 @@ final class ExecutableLinesFindingVisitor extends NodeVisitorAbstract
             return null;
         }
 
-        if ($node instanceof Node\Stmt\Interface_) {
+        if ($node instanceof Node\Stmt\Interface_ ||
+            $node instanceof Node\Attribute
+        ) {
             foreach (range($node->getStartLine(), $node->getEndLine()) as $line) {
                 $this->unsets[$line] = true;
             }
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
The proposed changes mark every line spanned by an attribute as non-executable. This makes sense to me and I will merge these changes.

This does not really explain the `--processes=1` vs `--processes=16` discrepancy, though. That smells like a static-analysis cache / merge-path difference where single-process likely hit a cached/short-circuit path that skipped attributes, while parallel runs went through the visitor. Once attributes are unconditionally unset, both paths converge, so the user-visible bug should disappear even though the underlying merging asymmetry isn't touched.

sebastianbergmann:
> releases are needed also for `13` and `14` branches

I have released [phpunit/php-code-coverage 12.5.5](https://github.com/sebastianbergmann/php-code-coverage/releases/tag/12.5.5) and [phpunit/php-code-coverage 14.1.1](https://github.com/sebastianbergmann/php-code-coverage/releases/tag/14.1.1) with this fix, but there will be no more releases for the 13.x version series.

Slamdunk:
> This does not really explain the `--processes=1` vs `--processes=16` discrepancy, though. That smells like a static-analysis cache / merge-path difference where single-process likely hit a cached/short-circuit path that skipped attributes, while parallel runs went through the visitor. Once attributes are unconditionally unset, both paths converge, so the user-visible bug should disappear even though the underlying merging asymmetry isn't touched.

Yeah: one process lets PHP decide the executable lines of `MyClassWithAttributes` because it uses it, while the other process has to leverage `ExecutableLinesFindingVisitor` because it doesn't use it.

PR #1153, opened 2026-04-22
Filter HTML code coverage report by test size

sebastianbergmann:
This adds an interactive test-size filter to the HTML code coverage report. Every directory and file page now shows a button group above the coverage table with **Small**, **Medium**, and **Large** toggles and an **All** reset button.

Selecting one or more sizes recalculates all coverage numbers on the page to show only the coverage achieved by tests of the selected size(s).

PHPUnit already records the size of each test (declared via the `#[Small]`, `#[Medium]`, and `#[Large]` attributes) in the coverage data; the changes proposed here put that information to use in the report.

## Motivation

A class may show 100% coverage overall while only a fraction of it is exercised by small (unit) tests, with the rest reached only through large end-to-end tests. The filter makes this visible directly in the report: toggle **Small** and you see what your unit tests alone cover.

## What the user sees

* A **Small / Medium / Large / All** button group on directory and file pages, in all report variants (line, branch, and path coverage). **All** is active by default and shows the unfiltered report.
* **Small**, **Medium**, and **Large** are checkboxes, so any combination can be selected (e.g. Small + Medium).
* When a filter is active, the progress bars, percentages, and "tested / total" counts update in place for lines, methods/functions, and classes/traits, on both directory pages (per directory and per file) and on file pages (per class, trait, method, and function). The red/yellow/green thresholds are applied to the recalculated percentages.
* Under a filter, a method counts as tested only if tests of the selected sizes alone fully cover it, and a class or trait counts as tested only if all of its methods do.

### "All" vs. "Small + Medium + Large"

These are intentionally not equivalent. **All** ignores test sizes entirely and counts every covered line. **Small + Medium + Large** counts only lines covered by at least one test with a declared size. Lines covered exclusively by tests of unknown size (no size attribute) count as uncovered under any filter. The difference between the two views shows how much coverage comes from tests without a declared size.

### Screenshots

grafik
grafik
grafik
grafik


sebastianfeldmann:
Please switch the small medium large buttons to checkboxes, then it's easier to handle different combinations even for future sizes you might introduce.

sebastianbergmann:
> Please switch the small medium large buttons to checkboxes, then it's easier to handle different combinations even for future sizes you might introduce.

Implemented. Thank you for the suggestion!

PR #1186, opened 2026-05-20
Make `Merger::merge()` accept any kind of iterable

Jean85:
This adds a refactor to the `Merger` class, allowing async merge of serialized coverage results (edit: by leveraging generators), for Paraunit (and similar tools) .

This would help me with https://github.com/facile-it/paraunit/pull/390:
 * allows to merge incoming coverage results just-in-time, and not in bulk at the end of tests execution
 * it has memory advantages, because it allows me to immediately erase the serialized coverage file as soon as it's merged
 * it doesn't change the logic or structure of the method, just switches from `array` (list) to `\Generator`
 * I maintained backward compatibility without code duplication, just by converting the array into a generator internally

### DIFF
--- a/src/Serialization/Merger.php
+++ b/src/Serialization/Merger.php
@@ -12,7 +12,6 @@ namespace SebastianBergmann\CodeCoverage\Serialization;
 use function array_key_exists;
 use function array_merge;
 use DateTimeImmutable;
-use Generator;
 use SebastianBergmann\CodeCoverage\Version;
 
 /**
@@ -23,7 +22,7 @@ use SebastianBergmann\CodeCoverage\Version;
 final readonly class Merger
 {
     /**
-     * @param list<non-empty-string> $paths
+     * @param iterable<non-empty-string> $paths
      *
      * @throws DriverMismatchException
      * @throws EmptyPathListException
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
~~Is the `asyncMerge()` method supposed to be `public`?~~

Nevermind; sorry for the noise.

Jean85:
Yes, it's definitely intended :D I would like to use it in Paraunit.

I've added a dedicated test for the new method.

sebastianbergmann:
A couple of notes:

- I think the name `asyncMerge()` is misleading. There is nothing asynchronous here; generators are lazy/streaming, and the merge still runs synchronously in a sequential loop. Since this becomes a public method covered by the backward-compatibility promise, I would rather not ship a name that will confuse people for the lifetime of the API. Something like `mergeStream()` / `mergeIterable()` / `mergeLazy()` describes what it actually does.

- The body only needs to iterate, so requiring `Generator` specifically is more restrictive than necessary. It rejects plain `Iterator`/`IteratorAggregate` implementations a caller might already have. If the method takes `iterable<non-empty-string>` and the body uses a `foreach` with a "first item" flag rather than manual `valid()`/`current()`/`next()` calls, then `merge()` can delegate directly (an array is `iterable`), and the `createGeneratorFor()` helper disappears. Right now it only exists to wrap an array into a generator so it can be iterated back out as a generator, which is a round-trip we can avoid.

- Please also cover the distinct behavior of the new entry point with tests. At minimum an empty generator/iterable producing `EmptyPathListException` (the `!valid()` branch), and ideally one case asserting merged content through the new method.

Jean85:
@sebastianbergmann your comments made me think that I could try a different approach: [157f66b](https://github.com/sebastianbergmann/php-code-coverage/pull/1186/commits/157f66bd5bec0d805879759b38eece1da700199b)

Since `iterable` includes `array`, I could just widen the method signature, without adding a new method. This means also that all previous tests apply, and I just left the one that explicitly tests using a generator.

Only drawback, I couldn't find a clean way to do the same as `array_slice` with a generic `iterable`, so I had to be somewhat "creative" 😅 

Let me know if this is acceptable to you.

sebastianbergmann:
Not sure what you mean by "creative", but can we get rid of the `??=` operators? Thank you.

Jean85:
> Not sure what you mean by "creative", but can we get rid of the `??=` operators? Thank you.

That's exactly what I meant by "creative" 😅  PHPStan complains about "possibly unassigned/null variables" because it considers the case where the iterable is empty and the loop doesn't run; that's caught by the `EmptyPathListException` check, but PHPStan doesn't know about that.

I've found in 8f2d548 a different approach to solve this, but I had to use a couple of `assert`s to silence some PHPStan issues; I did it because [I couldn't reproduce the issue to report it](https://phpstan.org/r/7d20833b-6624-45c1-aeef-642bc52fc0fd) so I expect it to be solved in future PHPStan releases.

Let me know if now it's ok for you.

PR #1212, opened 2026-06-01
Filenames in HTML report are not escaped

edorian:
Hi,

I've recently been playing around with oddly named files as part of a testing project, and while using the code coverage report, I’ve noticed that they sometimes break the layout. 

The fix feels somewhat minimal, so hopefully this is the right place and a sensible change.


sebastianbergmann:
I backported the changes that apply to `12.5` to `12.5` and cherry-picked your changes into `14.1` from where I merged them to `main`. Thank you!

PR #1244, opened 2026-07-02
Do not build code unit map when test has no coverage targets

janedbal:
### Problem

- The `Test\Target\Mapper` is built even when no test in the entire suite uses `#[CoversClass]` / `#[UsesClass]`.
- Building the mapper means `MapBuilder::build()` statically analyses **every file in the filter**, so a suite that requests any coverage report pays a full source-tree analysis per PHP process, even though the resulting map is never consumed.
- PHPUnit passes an empty `TargetCollection` to `CodeCoverage::validate()` / `append()` for every test that declares no coverage targets, and both call sites build the mapper before noticing the collection is empty.

Real-world numbers (app with **~24k files** under `<source>`, 5 fast unit tests, `--coverage-cobertura`, PCOV, cold cache):

  | | wall time | peak memory |
  | --- | --- | --- |
  | latest 12.5.7 | 53.5 s | 657 MB |
  | with this change | **10.6 s** | **426 MB** |

Without this patch, we are unable to migrate from PHPUnit 11 to 12+ as our overall CI time for tests inscreases from ~7 mins to ~12 mins.

### Notes:
  - Generated report is byte-identical. 
  - Projects that do use `#[CoversClass]`/`#[UsesClass]` are unaffected.

### DIFF
--- a/src/CodeCoverage.php
+++ b/src/CodeCoverage.php
@@ -259,10 +259,10 @@ final class CodeCoverage
         $linesToBeUsed    = [];
 
         if ($covers !== false) {
-            $linesToBeCovered = $this->targetMapper()->mapTargets($covers);
+            $linesToBeCovered = $covers->isEmpty() ? [] : $this->targetMapper()->mapTargets($covers);
         }
 
-        if ($linesToBeCovered !== false) {
+        if ($linesToBeCovered !== false && $uses->isNotEmpty()) {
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Merged manually, thank you.

janedbal:
[Ref](https://github.com/sebastianbergmann/php-code-coverage/commit/d342c699a8126fec1b769a8820552275a612f70a), thank you!

